Issue #14, opened 2018-10-28
Web support


Benny-:
Nyx supported web. This issue is about adding support for web environment back into Nyxx.

Benny-:
The following imports can't be used in the browser environment:

```dart
import 'dart:isolate';
import "dart:mirrors";
```

Benny-:
Closed with 84528a33513c33c0517548963186935f8ed17bb9

PR #16, opened 2018-10-30
Web support

Benny-:
See #14


lib/src/nyxx/Shard.dart
```diff
@@ -133,12 +131,12 @@ class Shard {
...
             "token": _client._token,
             "properties": <String, dynamic>{
-              "\$os": Platform.operatingSystem,
+              "\$os": operatingSystem,
               "\$browser": "nyxx",
               "\$device": "nyxx",
             },
             "large_threshold": _client._options.largeThreshold,
-            "compress": true
+            "compress": browser ? false : true
```
l7ssha: you can just do `!browser`

lib/src/nyxx/internal/Http.dart
```diff
@@ -285,8 +285,11 @@ class Http {
   Logger _logger = Logger.detached("Http");
 
   Http._new() {
-    this._headers= {'User-Agent':
-        'DiscordBot (https://github.com/l7ssha/nyxx, ${_Constants.version})'};
+    if(!browser)
```
l7ssha: "Clients using the HTTP API must provide a valid User Agent". [See](https://discordapp.com/developers/docs/reference#user-agent)

lib/src/nyxx/internal/Http.dart
```diff
@@ -296,12 +299,15 @@ class Http {
       bool beforeReady = false,
       Map<String, String> headers = const {},
       String reason}) async {
+    final Map<String, String> _headers = Map.from(this._headers)..addAll(headers);
+    if(!browser)
```
l7ssha: I think web client should be also able to set audit logs

lib/src/utils/Binder.dart
```diff
@@ -2,28 +2,6 @@ import 'package:nyxx/nyxx.dart';
...
-      for(var incl in classRefl.declarations.values.whereType<VariableMirror>()) {
-        if(meta.streamName == incl.simpleName.toString()) {
-          instanceThis.getField(incl.simpleName).reflectee.listen((evnt) {
-            Future.microtask(() => lib.invoke(decl.simpleName, [evnt]));
-          });
-        }
-      }
-    }
-  }
-}
```
l7ssha: Why did you remove this completely instead of moving to other place

Benny-: It was not referenced. Would you still like it somewhere? If so, where?

lib/src/nyxx/internal/Http.dart
Benny-: Setting that header was not allowed due to CORS.

lib/src/nyxx/internal/Http.dart
Benny-: Yikes! That is impossible for a web script. I will ask about this in discord api chat.

PR #26, opened 2018-11-17
Removed auto-sharding. Closes #21

### LINKED ISSUE #21: Replace internal sharding by user managed sharding
At the moment Nyxx has "internal sharding". It makes multiple connections to the discord gateway. But it all operates within the same process and does not help alleviate any problems related to processing power. This internal sharding does not allow Nyxx to process more messages.

This ticket is about replacing the current internal sharding and moving the responsibility to the caller of the library. The bot will have two extra optional parameters. The shard id and the maximum shard number. The user of the library must make sure to start all Nyxx shards.

Benny-:


l7ssha:
We can't just remove internal sharding - we should replace it with something else. I'm working on alternative solution for this problem.

PR #37, opened 2019-06-02
Cache return null if no value is found

### LINKED ISSUE #36: Can't send DM to user
When trying to send a DM to a user i get

```
Exception has occurred.
StateError (Bad state: No element)
```

Relevant code
```dart
e.guild.members[discord.Snowflake(id)].send(content: msg);
```

luaVolk:
In Dart `firstWhere` doesn't return null if no value was found, it throws a StateError instead unless `orElse` is specified.
This causes an error when trying to send a DM to a user because the channel isn't in the Cache. See https://github.com/l7ssha/nyxx/blob/development/lib/src/nyxx/objects/user/User.dart#L59

Fixes #36


l7ssha:
Thanks for contributing :D

PR #40, opened 2019-08-23
Fix misspell in postprocessor documentation

petrspelos:
- Postprocessor's description used to refer to it as a **Preprocessor**.


petrspelos:
So I found this one while playing around with the library. 😊 
It caught me off-guard a little.

PR #54, opened 2020-03-19
Fix #53 - CommandFramework improperly responds when a stream is passed

### LINKED ISSUE #53: CommandFramework improperly responds when a stream is passed
A stream is allowed to be set but when you set it, the code looks to ignore it and rather would raise an error saying that a framework cannot be made without a prefix - specifically the bit seen below

```dart 
client.onReady.listen((_) {
    if (prefix == null && stream == null) {
      prefix = client.self.mention;
      stream = client.onSelfMention;
    } else if (stream == null && prefix != null)
      stream = client.onMessageReceived;
    else {
      _logger.severe("CANNOT CREATE FRAMEWORK WITHOUT PREFIX");
      exit(1);
    }
```

I dunno how exactly the intended behaviour would be where if a stream was given and a prefix was not given. Logically I assume you would just inherit the prefix to be the bot's mention and the stream to be the given stream.

If that's the intended path of what you want to do, the final else could be changed to 
```dart
else if (stream != null && prefix == null) 
        prefix = client.self.mention; 
```
instead and not stop execution because in another scenario all that would be left is `stream != null && prefix != null` which would be the intended action in my scenario (because I want stream & prefix to be custom)

(also as always I'm on the rewrite_modular branch)

One-Nub:
Removes the final else & replaces it with the option to define a stream but not a prefix. By also removing the final else, the option can be made to pass both a prefix and a stream to the constructor and not have the bot shut down instantly.

### DIFF
--- a/nyxx.commands/lib/src/CommandsFramework.dart
+++ b/nyxx.commands/lib/src/CommandsFramework.dart
@@ -45,10 +45,8 @@ class CommandsFramework {
         stream = client.onSelfMention;
       } else if (stream == null && prefix != null)
         stream = client.onMessageReceived;
-      else {
-        _logger.severe("CANNOT CREATE FRAMEWORK WITHOUT PREFIX");
-        exit(1);
-      }
+      else if (stream != null && prefix == null)
+        prefix = client.self.mention;
 
       stream.listen((MessageEvent e) {
         if (ignoreBots && e.message.author.bot) return;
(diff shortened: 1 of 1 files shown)


nyxx.commands/lib/src/CommandsFramework.dart
```diff
@@ -45,10 +45,8 @@ class CommandsFramework {
         stream = client.onSelfMention;
       } else if (stream == null && prefix != null)
         stream = client.onMessageReceived;
-      else {
-        _logger.severe("CANNOT CREATE FRAMEWORK WITHOUT PREFIX");
-        exit(1);
-      }
+      else if (stream != null && prefix == null)
+        prefix = client.self.mention;
```
l7ssha: Please use curly braces around if

PR #55, opened 2020-03-21
Enhancing the potential usage of CommandsFramework by making dispatch() public.

One-Nub:
Most of my reasoning behind the changes are listed in the ~~mile long~~ description on commit https://github.com/l7ssha/nyxx/commit/de4e580c9aa7ebb828b22bd9f6f3a2c2d8329fad. (For clarity; a guild's stream is a stream that only contains message events for that guild)

The basic gist of it is it'll make it easier for users to make their own handler that triggers commands but still gets the features that CommandsFramework offers (command discovery, the metadata tags, etc...).

A potential flaw in this I see is that if a user defines a prefix but then tries to also manually dispatch events, the bot would still respond to the defined prefix when CommandsFramework is setup (assuming a stream isn't given). But if the user didn't define a prefix, the bot would use the mention which would allow for a custom prefix passed to dispatch and a constant trigger method that would be out of the way of conflicts with other bots.

I dunno if this is exactly a proper idea for the way the library works.

### DIFF
--- a/nyxx.commands/lib/src/CommandsFramework.dart
+++ b/nyxx.commands/lib/src/CommandsFramework.dart
@@ -45,11 +45,12 @@ class CommandsFramework {
   }
 
   ///Listens to messages sent that contain the prefix and triggers the dispatching of the command.
-  void _streamListener({Stream<MessageEvent> stream}) async {
+  void _streamListener({Stream<MessageEvent>? stream}) async {
     if (!client.ready) {
       //Allows to start listening if onReady has already been recieved
       await client.onReady.first;
     }
+
     if (prefix == null && stream == null) {
       prefix = client.self.mention;
       stream = client.onSelfMention;
@@ -59,9 +60,9 @@ class CommandsFramework {
       prefix = client.self.mention;
     }
 
-      stream!.listen((MessageEvent e) {
-        if (ignoreBots && e.message?.author != null && e.message!.author!.bot) return;
-        if (!e.message!.content.startsWith(prefix)) return;
+    stream!.listen((MessageEvent e) {
+      if (ignoreBots && e.message?.author != null && e.message!.author!.bot) return;
+      if (!e.message!.content.startsWith(prefix)) return;
 
       Future(() => dispatch(e));
     });
@@ -168,12 +169,17 @@ class CommandsFramework {
   }
 
   /// Dispatches onMessage event to framework.
-  Future _dispatch(MessageEvent e) async {
+  Future dispatch(MessageEvent e, {String? prefix}) async {
     if(e.message == null) {
       return;
     }
 
-    var cmdWithoutPrefix = e.message!.content.replaceFirst(prefix, "").trim();
+    String cmdWithoutPrefix;
+    if (prefix == null) {
+      cmdWithoutPrefix = e.message!.content.replaceFirst(this.prefix, "").trim();
+    } else {
+      cmdWithoutPrefix = e.message!.content.replaceFirst(prefix, "").trim();
+    }
 
     _CommandMetadata matchedMeta;
     try {
--- a/nyxx.commands/lib/src/CommandsFramework.dart
+++ b/nyxx.commands/lib/src/CommandsFramework.dart
@@ -23,12 +23,14 @@ class CommandsFramework {
 
   Nyxx client;
 
+  bool ignoreBots;
+
   /// Creates commands framework handler. Requires prefix to handle commands.
   CommandsFramework(this.client,
       {this.prefix,
       Stream<MessageEvent> stream,
       Duration roundupTime = const Duration(minutes: 2),
-      bool ignoreBots = true,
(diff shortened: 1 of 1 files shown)


l7ssha:
Can you provide example usage of this feature?

One-Nub:
An idea could be:
```dart
import 'package:nyxx/nyxx.dart';
import 'package:nyxx/Vm.dart';
import 'package:nyxx.commands/commands.dart';

Future<void> main() async {
  Nyxx bot = NyxxVm("legit token");
  CommandsFramework handler = CommandsFramework(bot)..discoverCommands();

  await bot.onReady.first;

  for (Snowflake guild in bot.guilds.keys) {
    var stream = await guildStream(guild, bot.onMessageReceived);
    //Imagine there's a proper way to get a guild's prefix here & isolates
    //all this instead of a for each loop
    String prefix;
    if(guild.id == 1)
      prefix = "1";
    else
      prefix = "2";

    stream.listen((e) {
      handler.dispatch(e, prefix: prefix);
    });
  }
}

Stream<MessageEvent> guildStream(Snowflake guildID,
  Stream<MessageEvent> source) async* {
  await for(MessageReceivedEvent msg in source) {
    if(msg.message.guild != null && msg.message.guild.id == guildID)
    {
      yield msg;
    }
  }
}

```

A better usage could be seen here (big work in progress be warned):
https://github.com/One-Nub/remindme-bot/blob/master/src/main.dart#L33
https://github.com/One-Nub/remindme-bot/blob/master/src/GuildHandlerUtils.dart#L19

l7ssha:
Okay seems cool to have feature like that. I will suspend this PR for some time because I just started working on NNBD in nyxx

l7ssha:
@One-Nub can you apply appropriate changes to changed api?

One-Nub:
Yea probably, I'll have to read up on the nndb syntax a bit more but then I should be able to

l7ssha:
What about command from DM's? Also I think api would look more consistent by enhancing CommandsFramework with optional callback which will provide guild (or null if dm (?)) or async generator based on mentioned before data. That approach should be also more resource efficient, esspecially memory, It would look like: 
```dart
import 'package:nyxx/nyxx.dart';
import 'package:nyxx/Vm.dart';
import 'package:nyxx.commands/commands.dart';

Future<bool> commandCallback(Guild? guild, String messageContent, MessageReceivedEvent event) async {
  if(guild == null && messageContent.startsWith("prefix1")) {
    return true;
  } else if(messageContent.startsWith("prefix2)) {
    return true;
  }

  return false;
}

Future<void> main() async {
  Nyxx bot = NyxxVm("legit token");
  CommandsFramework handler = CommandsFramework(bot, commandCallback)..discoverCommands();
```

But I don't know if it meets all your thoughs about that feature

One-Nub:
I mean quite honestly I'm having a hard time understanding where commandCallback would fit inside the api. So with passing commandCallback would there be something inside the constructor that would dispatch based off of if it was a guild or a DM? I suppose that idea appears to work if you want a separate prefix in the DMs (so a public prefix) but how would one go about telling the callback object what `prefix2` would be?

The example (I gave) above is a pretty isolated scenario where one would want a specific prefix in just one guild but if I wanted a specified prefix in multiple guilds I would have to have several callbacks(?).

One-Nub:
Ok it took a bit but I think I'm understanding more now your thoughts on how commandCallback would work. 

Just for clarification would that be something the user would try to implement and then pass to commandsFramework to use (basically your example?) & it would use that (assuming it just returns a bool) to determine if it should dispatch? 

That's pretty much the understanding of the idea I've had.

l7ssha:
I think it serves same or similar functionality for more consistent and readable API.
```dart
Future<bool> commandCallback(Guild? guild, String messageContent, MessageReceivedEvent event) async {
  if(guild == null && messageContent.startsWith("prefix1")) {
    return true;
  } else if(guild!.id == 67 && messageContent.startsWith("prefix2")) {
    return true;
  } else if(messageContent.startsWith("prefix2")) {
    return true;
  }

  return false;
}
```

l7ssha:
it could be even revered into your example

```dart
Future<bool> commandCallback(Guild? guild, String messageContent, MessageReceivedEvent event) async {
  String prefix;
  if(guild.id == 1) {
    prefix = "1";
  } else {
    prefix = "2";
  }
  
  if(messsageContent.startsWith(prefix)) {
    return true;
  }

  return false;
}
```

One-Nub:
Oh that's nice, I like that idea. Definitely would keep the api more readable.

l7ssha:
@One-Nub Would you like to implement this feature in that form or I can start working on this?

One-Nub:
You can go ahead and start on it

l7ssha:
Cool, all right. This PR will stay open as he is source of informations. I will created another PR with other version of this feature.

l7ssha:
Feature merged in #58.

PR #63, opened 2020-04-18
Various minor improvements

aayanraja210:
Made some minor improvements, nothing major. Overrided hashCode for "==" operator, temporarily changed all (as bool? ?? false) or (as int? ?? 0) to (as bool) or (as int) since bool? and int? was being casted to bool and int. Also changed _getSymbolName to use regex rather than replace ' " ( and )' with "".


l7ssha:
Also due deprecation of `CommandFramework` revert changes to that module. It will be removed when `Commander` will be merged into `rewrite_modular`

l7ssha:
Closing due no response to requested changes. If you want to merge these changes, feel free to reopen this PR.

PR #69, opened 2020-05-07
Fixing setPresence method

One-Nub:
Resolves both a serialization error to json & an issue regarding the packet format that discord expects for the presence.

### DIFF
--- a/nyxx/lib/src/Shard.dart
+++ b/nyxx/lib/src/Shard.dart
@@ -44,14 +44,14 @@ class Shard implements Disposable {
   /// Allows to set presence for current shard.
   void setPresence({UserStatus? status, bool? afk, Activity? game, DateTime? since}) {
     var packet = <String, dynamic> {
-      if(status != null) 'status' : status,
-      if(afk != null) 'afk' : afk,
+      'status' : (status != null) ? status._value : UserStatus.online._value,
+      'afk' : (afk != null) ? afk : false,
       if(game != null) 'game' : <String, dynamic> {
         'name' : game.name,
-        'type' : game.type,
-        if(game.type == ActivityType.streaming) 'url' : game.url
+        'type' : game.type._value,
+        if(game.type._value == ActivityType.streaming._value) 'url' : game.url
       },
-      if(since != null) 'since' : since.millisecondsSinceEpoch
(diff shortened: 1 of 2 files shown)


nyxx/lib/src/Shard.dart
```diff
@@ -44,14 +44,14 @@ class Shard implements Disposable {
...
-      if(status != null) 'status' : status,
-      if(afk != null) 'afk' : afk,
+      'status' : (status != null) ? status._value : UserStatus.online._value,
+      'afk' : (afk != null) ? afk : false,
       if(game != null) 'game' : <String, dynamic> {
         'name' : game.name,
-        'type' : game.type,
-        if(game.type == ActivityType.streaming) 'url' : game.url
+        'type' : game.type._value,
+        if(game.type._value == ActivityType.streaming._value) 'url' : game.url
```
l7ssha, line 52: Can you remove `_value` and modify `==` operator in `ActivityType`
```dart
other is ActivityType && other._value == this._value
```

nyxx/lib/src/Shard.dart
```diff
@@ -44,14 +44,14 @@ class Shard implements Disposable {
   /// Allows to set presence for current shard.
   void setPresence({UserStatus? status, bool? afk, Activity? game, DateTime? since}) {
     var packet = <String, dynamic> {
-      if(status != null) 'status' : status,
-      if(afk != null) 'afk' : afk,
+      'status' : (status != null) ? status._value : UserStatus.online._value,
```
l7ssha, line 47: Call to `toString()` would be probably better than accessing private property.

l7ssha:
All "enums" should probably have something like `getValue` getter because calling _value isn't good practice. But this is for another issue.

nyxx/lib/src/Shard.dart
One-Nub: Should I also remove it for the moment in `game.type._value`? The only workaround I can think of that would leave it in a usable state would be to parse an int from the toString of game.type ultimately looking like: 
```dart
'type' : int.parse(game.type.toString())
```

l7ssha: In that case not because we need int as in documentation. This is related to #70 that we need sometimes raw value of enum and now we don't have way to.

PR #76, opened 2020-05-20
Dont require cache when possible

l7ssha:
Due discord shard intents feature, `Nyxx` have to be able to operate without any state or cache whatsoever. 
Internal structure of objects have to be reworked in order to achieve that:
- channels internal structure changed completely in order to meet requirements:
   * Removed `GuildChannel` mixin
   * Added VoiceChannel mixin
   * MessageChannel is now mixin
   * Added CachelessGuildChannel and CacheGuildChannel
- *TODO*

Events are already stateless and don't require cache to be emitted (6fbd1f1).

##### TODO: 
- [ ] events
- [ ] channels 
- [ ] guild
- [ ] message
  - [ ] role
  - [ ] emoji

Related #611


l7ssha:
Current Channel class hierarchy:
nyxx-class-hierachy

PR #110, opened 2021-03-11
Use entity.isEntityName instead of entity.name

WasserEsser:
Currently, aliases don't work because messageParts.first is only checked against entity.name.
Using entity.isEntityName fixes this.


l7ssha:
Thanks for contributing @WasserEsser :)

l7ssha:
Verison with your fix was just pushed to pub @WasserEsser

WasserEsser:
Thanks!

Issue #111, opened 2021-03-12
context.getArguments() breaks with command aliases


WasserEsser:
context.getArguments() currently only works with the default command name.
This is because it removes the prefix + fullCommandName from the message and takes the rest as arguments. With aliases however, there could be multiple combinations of fullCommandName.

Think of a CommandGroup with name 'hello' and an alias of 'h', with a subCommandEntity with the name 'world' and the alias 'w'.

This would mean that the command can be triggered by 'hello world', 'hello w', 'h world' and 'h w', but getting the arguments only works in the case of 'hello world'.

l7ssha:
all right, I see problem here. Will take a look today :)

l7ssha:
Fixed with 79035b3. Thanks for reporting issue :)

PR #113, opened 2021-03-12
Match final command name using RegEx to fix argument extraction

WasserEsser:
The recent commit to fix the initial issue #111 wasn't working for me and introduced all sorts of problems with getting command arguments. This PR is trying to fix the issue once and for all. 

I have introduced a new getter and replaced getFullCommandName() with getFullCommandMatch().
This function returns a RegEx match which includes all possible command names.
By using RegEx, we are no longer in the dark on what the final command is and CommandContext.getArguments() is going to remove the actual command accordingly and grab only the arguments.

An example for the scenario would be a CommandGroup with name `quote` and and alias of `q`.
This CommandGroup has several subcommands, one of which is `remove` with an alias of `rm`.

The messages that can trigger this command are the following:

- quote remove
- q remove
- quote rm
- q rm

Using isEntityName, we're already matching the correct name for command execution, but we were still passing an incorrect command name to the CommandContext, which would result in the actual command not being replaced and the argument list containing parts of the command.

### DIFF
--- a/nyxx.commander/lib/src/Commander.dart
+++ b/nyxx.commander/lib/src/Commander.dart
@@ -105,15 +105,9 @@ class Commander with ICommandRegistrable {
     // Example: (?<finalCommand>(quote|q) (remove|rm))
     // This will match the command `quote remove`, `q remove`, `quote rm` and `q rm`
 
-    final match = RegExp("(?<finalCommand>${matchingCommand.getFullCommandMatch()})").firstMatch(
-      event.message.content,
-    );
+    final match = RegExp("(?<finalCommand>${matchingCommand.getFullCommandMatch()})").firstMatch(event.message.content);
 
     final finalCommand = match?.namedGroup("finalCommand");
-
-    if (finalCommand == null) {
-      return;
-    }
 
     // construct commandcontext
     final context = CommandContext._new(
--- a/nyxx.commander/lib/src/Commander.dart
+++ b/nyxx.commander/lib/src/Commander.dart
@@ -102,8 +102,8 @@ class Commander with ICommandRegistrable {
 
     // Builds a RegEx that matches the full command including their parents and all possible
     // aliases of the final command entity and their parents.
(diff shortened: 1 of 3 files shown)


nyxx.commander/lib/src/Commander.dart
```diff
@@ -94,22 +94,34 @@ class Commander with ICommandRegistrable {
...
 
-    final fullCommandName = _getFullCommandFromMessage(matchingCommand, messageContentParts);
+    // Builds a RegEx that matches the full command including their parents and all possible
+    // aliases of the final command entity and their parents.
+    // Example: (?<finalCommand>(quote|quotes) (remove|rm))
+    // This will match the command `quote remove`, `quotes remove`, `quote rm` and `quotes rm`
+
+    final match = RegExp("(?<finalCommand>${matchingCommand.getFullCommandMatch()})").firstMatch(
+      event.message.content,
+    );
```
l7ssha, line 110:
```suggestion
    final match = RegExp("(?<finalCommand>${matchingCommand.getFullCommandMatch()})").firstMatch(event.message.content);
```

nyxx.commander/lib/src/Commander.dart
```diff
@@ -94,22 +94,34 @@ class Commander with ICommandRegistrable {
...
+
+    final finalCommand = match?.namedGroup("finalCommand");
+
+    if (finalCommand == null) {
+      return;
+    }
```
l7ssha, line 116: Is that check really needed? Handler will stop executing higher I think we won't need that.

l7ssha (changes requested):

nyxx.commander/lib/src/Commander.dart
WasserEsser: In practice, finalCommand should never be null because a valid command was found using _findMatchingCommand, so the RegEx should also be able to figure out the full command.

I have removed the check.

l7ssha:
Thanks for contributing. Will push changes soon to pub.

PR #119, opened 2021-05-11
Commander : commander is  not case sensitive now.

SushantChandla:


l7ssha:
Why would you want case sensitive commands? Why that should be marged? If that should be merged maybe better would be to provide option to enable/disable this feature?

SushantChandla:
> Why would you want case sensitive commands?

I don't want it to be case sensitive.
It was case sensitive, so i have fixed that.

nyxx_commander/lib/src/CommandHandler.dart
```diff
@@ -19,7 +19,7 @@ abstract class CommandEntity {
   CommandEntity? get parent;
 
   /// A list of valid command names
-  List<String> get commandNames => [if (this.name.isNotEmpty) this.name, ...this.aliases];
+  List<String> get commandNames => [if (this.name.isNotEmpty) this.name.toLowerCase(),]+aliases.map((e) => e.toLowerCase()).toList();
```
l7ssha, line 22: `...aliases.map((e) => e.toLowerCase()).toList()` would be better, but i'm not sure if that's possible

SushantChandla: Updated.

HarryET (approved):
Looks good :)

PR #128, opened 2021-05-29
Added message to component interaction

HarryET:
Added the message cacheable to the interaction for component.

### DIFF
--- a/nyxx/lib/src/utils/EntityUtility.dart
+++ b/nyxx/lib/src/utils/EntityUtility.dart
@@ -114,4 +114,16 @@ class EntityUtility {
   static GuildMessage createGuildMessage(
       INyxx client, Map<String, dynamic> rawJson) =>
       GuildMessage._new(client, rawJson);
+
+  /// Creates a Guild Member object, can be used for other classes where you have correct rawJson data from the API.
+  /// ```dart
+  /// void main() {
+  ///     var bot = Nyxx("TOKEN");
+  ///     Map<String, dynamic> rawJson = /* rawJson from the API */
+  ///     DMChannel dmChannel = EntityUtility.createGuildMessage(bot, rawJson);
+  /// }
+  /// ```
+  static Message createMessage(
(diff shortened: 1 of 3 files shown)


l7ssha (changes requested):
Since interactons are available in DMs implementation needs to take that into consideration

l7ssha (approved):

nyxx/lib/src/utils/EntityUtility.dart
```diff
@@ -102,4 +102,28 @@ class EntityUtility {
...
+  /// ```dart
+  /// void main() {
+  ///     var bot = Nyxx("TOKEN");
+  ///     Map<String, dynamic> rawJson = /* rawJson from the API */
+  ///     DMChannel dmChannel = EntityUtility.createGuildMessage(bot, rawJson);
+  /// }
+  /// ```
+  static GuildMessage createGuildMessage(
+      INyxx client, Map<String, dynamic> rawJson) =>
+      GuildMessage._new(client, rawJson);
```
l7ssha, line 116: its pretty much useless if there is createMessage

nyxx_interactions/pubspec.yaml
```diff
@@ -12,3 +12,7 @@ environment:
 dependencies:
   logging: "^1.0.1"
   nyxx: "2.0.0-rc.5"
+
+dependency_overrides:
+  nyxx:
+    path: "../nyxx"
```
l7ssha, line 18: we dont need that

l7ssha (changes requested):

nyxx_interactions/pubspec.yaml
HarryET: done

nyxx/lib/src/utils/EntityUtility.dart
HarryET: done

l7ssha (approved):

Issue #152, opened 2021-07-17
No way to fetch Roles.


Technicolor-creamsicle:
**Is your feature request related to a problem? Please describe.**
When trying to manage a role on a server or count how many users are in the role there is no way to ensure that the bot has the role in the cache.

**Describe the solution you'd like**
A `fetchRole()` method to be added.

**Describe alternatives you've considered**
None.

**Additional context**
N/A

HarryET:
Will do, @l7ssha assign me please

l7ssha:
Fixed in #153

PR #153, opened 2021-07-18
Add Guild#fetchRoles

HarryET:
Add ability to fetch a guilds roles, no endpoint to fetch single role. Could be added as a method but leaving up to user to sort on the stream. Sorts issue #152

### DIFF
--- a/nyxx/lib/src/core/guild/Guild.dart
+++ b/nyxx/lib/src/core/guild/Guild.dart
@@ -287,6 +287,11 @@ class Guild extends SnowflakeEntity {
   Future<GuildSticker> fetchSticker(Snowflake id) =>
       client.httpEndpoints.fetchGuildSticker(this.id, id);
 
+  /// Fetches all roles that are in the server.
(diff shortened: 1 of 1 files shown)


l7ssha (approved):

nyxx/lib/src/core/guild/Guild.dart
```diff
@@ -287,6 +287,11 @@ class Guild extends SnowflakeEntity {
   Future<GuildSticker> fetchSticker(Snowflake id) =>
       client.httpEndpoints.fetchGuildSticker(this.id, id);
 
+  /// Fetches all roles that are in the server.
+  /// To get a single file use .firstWhere()
```
l7ssha, line 291: I dont think that information is needed here, fist line would be sufficient

l7ssha (changes requested):

nyxx/lib/src/core/guild/Guild.dart
HarryET: All sorted in latest push

Issue #156, opened 2021-07-26
Allow to skip syncing interactions commands and permissions


l7ssha:
**Is your feature request related to a problem? Please describe.**

Interactions should allow skipping sync of commands and permissions to avoid unnecessary api requests.

**Describe the solution you'd like**
Implement interface that will decide if syncing should or should not be skiped

**Additional context**
Pseudo code:

```dart
abstract class ICommandsSync {
  FutureOr<bool> shouldSyncCommands(Interable<CommandBuilder> commands); // or whatevet that builder is called
  FutureOr<bool> shouldSyncCommandPermisions(Interable<CommandBuilder> commands);
}

class LockFileCommandsSync implements ICommandsSync {
  // here you impl of lock file
}

class ManualCommandsSync implements ICommandsSync {
  ManualCommandsSync(this.syncCommands, this.syncPermissions);

  FutureOr<bool> shouldSyncCommands() => this.syncCommands;
  FutureOr<bool> shouldSyncCommandPermisions() => this.syncPermissions;
}

Interactions()
  ..syncOnReady(commandsSync: LockFileCommandsSync())
```

HarryET:
Addressed with PR #159

l7ssha:
Maybe it should also support #168?

HarryET:
Yup, could do.

HarryET:
This initial PR has been implemented in #159, And #168 should/might be implemented in #173.

PR #159, opened 2021-07-26
Add sync control

HarryET:
Addresses issue #156


nyxx_interactions/lib/src/interfaces/ICommandsSync.dart
```diff
@@ -0,0 +1,7 @@
+part of nyxx_interactions;
```
l7ssha, line 1: File should be moved to dir like `internal/sync`. I don't like structures like here

nyxx_interactions/lib/src/models/LockFileCommandSync.dart
```diff
@@ -0,0 +1,74 @@
+part of nyxx_interactions;
```
l7ssha, line 1: File should be moved to dir like internal/sync. I don't like structures like here

nyxx_interactions/lib/src/models/LockFileCommandSync.dart
```diff
@@ -0,0 +1,74 @@
...
+
+    if (other.permissionType != this.permissionType ||
+        other.permissionEntityId != this.permissionEntityId ||
+        other.permissionsGranted != this.permissionsGranted) {
+      return false;
+    }
+
+    return true;
+  }
+}
```
l7ssha, line 74: Implementation should be based on md5 (or something else, maybe there is something in standard lib) I think. You should create for example string with all meaningful params that could be changed and then hash with it so you will be short hash string that would be different each time something is changed

l7ssha (changes requested):

nyxx_interactions/lib/src/models/LockFileCommandSync.dart
```diff
@@ -0,0 +1,74 @@
...
+  final bool defaultPermissions;
+  final Iterable<_LockfilePermission> permissions;
+
+  _LockfileCommand(this.name, this.guild, this.defaultPermissions, this.permissions);
+
+  @override
+  bool operator ==(Object other) {
+    if (other is! _LockfileCommand) {
+      return super == other;
+    }
```
l7ssha, line 40: thats not needed here, just return false

nyxx_interactions/lib/src/models/LockFileCommandSync.dart
l7ssha: You also forgot to serialize options and descriptions, which are crucial

nyxx_interactions/lib/src/interfaces/ICommandsSync.dart
HarryET: 👌🏼

nyxx_interactions/lib/src/models/LockFileCommandSync.dart
HarryET: 👌🏼

nyxx_interactions/lib/src/internal/sync/LockFileCommandSync.dart
```diff
@@ -0,0 +1,123 @@
...
+        lockFileMapData[c.name] = new _LockfileCommand(
+          c.name,
+          c.description,
+          c.guild,
+          c.defaultPermissions,
+          c.permissions?.map((p) => new _LockfilePermission(p._type, p.id, p.hasPermission)) ?? [],
+          c.options.map((o) => new _LockfileOption(o.type.value, o.name, o.description, o.options ?? [])),
+        ).generateHash();
+      },
+    );
```
l7ssha, line 23: cant be for (final command in commands) be used?

HarryET: Yup that works too. Changes done 👍🏼

nyxx_interactions/lib/interactions.dart
```diff
@@ -1,40 +1,39 @@
 library nyxx_interactions;
```
l7ssha, line 1: have you check if whole thing is working properly? By changing commands, only perms and perms with commands?
Most of the api isn't public so I suppose we cannot test that?

l7ssha (changes requested):

nyxx_interactions/lib/interactions.dart
HarryET: I'm going to test once #172 has been fixed. As that may be an issue during testing

nyxx_interactions/lib/src/Interactions.dart
```diff
@@ -78,91 +82,78 @@ class Interactions {
...
-        ._client
-        .httpEndpoints
-        .sendRawRequest("/applications/${this._client.app.id}/commands/permissions", "PUT", body: globalBody);
+    final groupedGuildCommands = _groupSlashCommandBuilders(commandPartition.last);
+
+    final globalBody =
+        globalCommands.where((builder) => builder.permissions != null && builder.permissions!.isNotEmpty).toList();
+
+    if (globalBody.isNotEmpty) {
+      throw "Only guild commands may have permission's set. If you need custom permissions based on guild then you can send HTTP requests.";
```
l7ssha, line 93: thats a bad practise. we should only throw Error or Exception instances. and in case of future you should return Future.error. You can return there `StateError` with same message

nyxx_interactions/lib/src/Interactions.dart
```diff
@@ -78,91 +82,78 @@ class Interactions {
...
     });
   }
 
   /// Syncs command builders with discord immediately.
   /// Warning: Client could not be ready at the function execution.
   /// Use [syncOnReady] for proper behavior
-  Future<void> sync() async {
+  Future<void> sync({ICommandsSync syncRule = const ManualCommandSync()}) async {
+    final syncCommands = await syncRule.shouldSync(this._commandBuilders);
+    if (!syncCommands) {
```
l7ssha, line 136:
```suggestion
    if (!await syncRule.shouldSync(this._commandBuilders)) {
```
thats up to you

l7ssha (changes requested):

l7ssha (dismissed):

l7ssha (approved):

PR #238, opened 2021-11-01
Allow handlers on different nesting layers, closes #233

abitofevrything:
# Description

Previously handlers could not be registered for both subcommands and sub-subcommands because of the way preventing root handlers on slash commands with subcommands was implemented. This PR reworks that system to allow both subcommands and sub-subcommands to have handlers, while still preventing slash commands with subcommands having roor handlers.

## Type of change

# Checklist:

- [x] Ran `dartanalyzer --options analysis_options.yaml .`

### DIFF
--- a/nyxx_interactions/lib/src/Interactions.dart
+++ b/nyxx_interactions/lib/src/Interactions.dart
@@ -241,6 +241,8 @@ class Interactions {
   void _assignCommandToHandler(SlashCommandBuilder builder, SlashCommand command) {
     final commandHashPrefix = "${command.id}|${command.name}";
 
+    var allowRootHandler = true;
+
     final subCommands = builder.options.where((element) => element.type == CommandOptionType.subCommand);
     if (subCommands.isNotEmpty) {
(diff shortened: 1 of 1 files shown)


HarryET:
Thanks @abitofevrything for this PR. I will review it in more detail tomorrow, was this a flaw in design or an API change?

abitofevrything:
Hi, thanks for the quick reply. I think this was a flaw in design as Discord has always allowed slash commands and slash command groups on the same layer, but handlers would not be registered on both as explained above.
This is a small fix that shouldn't affect the API or any applications using `nyxx_interactions`.

abitofevrything:
Also concerning the repo change - I had not noticed that nyxx_interactions had its own repo, would you like me to close this and move it there?

l7ssha:
> Also concerning the repo change - I had not noticed that nyxx_interactions had its own repo, would you like me to close this and move it there?

PR should be moved there, and I will backport fix to next brach

abitofevrything:
WIll open an issue in the interactions repo.

PR #248, opened 2021-12-06
[feature] Guild events

l7ssha:
# Description

Guild events implementation. ~Subclass for each event type:~ (everything will be in one class) 

TODO:
 - [x] Intents
 - [x] New gateway events

## Type of change

# Checklist:

- [x] Ran `dartanalyzer --options analysis_options.yaml .`

### DIFF
--- a/lib/src/client_options.dart
+++ b/lib/src/client_options.dart
@@ -134,6 +134,9 @@ class GatewayIntents {
   /// Includes events: `TYPING_START`
   static const int directMessageTyping = 1 << 14;
 
+  /// Includes events: `GUILD_SCHEDULED_EVENT_CREATE`, `GUILD_SCHEDULED_EVENT_DELETE`, `GUILD_SCHEDULED_EVENT_UPDATE`, `GUILD_SCHEDULED_EVENT_USER_ADD`, `GUILD_SCHEDULED_EVENT_USER_REMOVE`
+  static const int guildScheduledEvents = 1 << 16;
+
   /// All unprivileged intents
   static const int allUnprivileged = guilds |
       guildBans |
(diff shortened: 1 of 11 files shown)


lib/src/client_options.dart
```diff
@@ -134,6 +134,9 @@ class GatewayIntents {
   /// Includes events: `TYPING_START`
   static const int directMessageTyping = 1 << 14;
 
+  /// Includes events: `GUILD_SCHEDULED_EVENT_CREATE`, `GUILD_SCHEDULED_EVENT_DELETE`, `GUILD_SCHEDULED_EVENT_UPDATE`, `GUILD_SCHEDULED_EVENT_USER_ADD`, `GUILD_SCHEDULED_EVENT_USER_REMOVE`
```
l7ssha, line 137: I'm not adding last to events beacuse they are experimental and I don't want to deal with it later. Will add if they become stable

lib/src/events/guild_events.dart
```diff
@@ -411,3 +412,42 @@ class GuildStickerUpdate implements IGuildStickerUpdate {
...
+  late final IGuildEvent event;
+
+  GuildEventUpdateEvent(RawApiMap raw, INyxx client) {
+    event = GuildEvent(raw['d'] as RawApiMap, client);
+  }
+}
+
+abstract class IGuildEventDeleteEvent {
+  IGuildEvent get event;
+}
```
l7ssha, line 444: all the events are looking the same, maybe we should just provide one call for all of them? Or do inheritance maybe in case they extends event data in future?

HarryET (approved):

lib/src/events/guild_events.dart
l7ssha: gonna leave as is for now :)

PR #288, opened 2022-01-16
feature: interaction-modals necessary changes

l7ssha:
# Description

This PR adds necessary fixes and changes to make modals feature possible in nyxx_interactions possible
 
## Type of change

# Checklist:

### DIFF
--- a/lib/src/core/message/components/component_style.dart
+++ b/lib/src/core/message/components/component_style.dart
@@ -1,6 +1,7 @@
 import 'package:nyxx/src/utils/enum.dart';
 
 /// Style for a button.
+@Deprecated("Use ButtonStyle")
 class ComponentStyle extends IEnum<int> {
   /// A blurple button
   static const primary = ComponentStyle._create(1);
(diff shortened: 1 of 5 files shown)


lib/src/core/message/components/component_style.dart
```diff
@@ -1,6 +1,7 @@
 import 'package:nyxx/src/utils/enum.dart';
 
 /// Style for a button.
+@Deprecated("Use ButtonStyle")
```
abitofevrything, line 4: Why mark as deprecated if a major version change is happening? Just removing it would be cleaner.

abitofevrything (changes requested):

lib/src/core/message/components/component_style.dart
l7ssha: thats a mistake, i mean major version increment since these changes are not breaking if not necessary this PR will be released as minor version. Im still targeting next since no one knows when modals gonna release. Will fix that version constraint

abitofevrything (dismissed):

lib/src/core/message/components/component_style.dart
l7ssha: done

PR #305, opened 2022-03-01
Patch emojis & new properties

Rapougnac:
# Description
My primary motivation was to fix bugs (e.g: `toList` not called on `map`, cannot react with custom emojis...).
After that, I saw there wasn't a lot of properties for emojis, like; theirs names was always `nyxx`, the `cdnUrl` was always returning static pngs even for animated emojis, and so on. So I basically fixed bugs and added new methods/properties described below.

## Type of change

- Added example for emojis

		- Defaut value of http for `CachePolicyLocation` is `true` instead of `false` (according to [the doc](https://github.com/nyxx-discord/nyxx/blob/68ebcde5c7fdf37a5dacb71777c02de988ad9afc/lib/src/internal/cache/cache_policy.dart#L27))
		- `encodeForAPI` in `IBaseGuildEmoji` works properly now

		- Added `fetchCreator` method to emoji
		- Added `resolve` method on `IResolvableGuildPartialEmoji` to return a `IGuildEmoji`

	   - `cdnUrl` is no more a property

# Checklist:


lib/src/core/message/guild_emoji.dart
```diff
@@ -102,19 +155,43 @@ class GuildEmoji extends BaseGuildEmoji implements IGuildEmoji {
...
   @override
   bool get isPartial => false;
 
+  /// Returns cdn url to emoji
+  @override
+  String get cdnUrl => "https://cdn.discordapp.com/emojis/$id.${animated ? "gif" : "png"}";
+
+  /// The user that created this emoji, not present if [fetchCreator] has not been called
+  @override
+  late Cacheable<Snowflake, IUser>? creator;
```
abitofevrything, line 171: The `late` keyword here isn't necessary.

example/emojis.dart
```diff
@@ -0,0 +1,34 @@
...
+  });
+  // This event is called when a message is received
+  bot.eventsWs.onMessageReceived.listen((event) {
+    if(event.message.content == '!emoji') {
+      final emoji = bot.emojis.values.firstWhere((emo) => emo.name == 'some_emoji');
+      final msg = await event.message.channel.sendMessage(MessageBuilder.content('Look at this emoji: $emoji'));
+      msg.createReaction(emoji);
+      // For unicode emoji use `UnicodeEmoji` class
+      msg.createReaction(UnicodeEmoji('🤔'));
+    }
```
abitofevrything, line 21: This function body needs to be marked as `async` for the await to work.

example/emojis.dart
```diff
@@ -0,0 +1,34 @@
...
+  });
+
+  // This event is called when a reaction has been added to a message
+  bot.eventsWs.onMessageReactionAdded.listen((event) {
+    final emoji = (event.emoji as IGuildEmojiPartial).isPartial ? (event.emoji as IGuildEmojiPartial).resolve() : event.emoji;
+    if(emoji is UnicodeEmoji) {
+      event.message?.channel?.sendMessage(MessageBuilder.content('Woah! This is a unicode emoji: ${event.emoji}'));
+    }
+    if((emoji as IGuildEmoji).name == 'some_emoji') {
+      event.message.channel.sendMessage(MessageBuilder.content('Woah! This is a custom emoji: $emoji'));
```
abitofevrything, line 31: `message` is nullable so needs a null-aware accessor (`?.`).

example/emojis.dart
```diff
@@ -0,0 +1,34 @@
...
+      // For unicode emoji use `UnicodeEmoji` class
+      msg.createReaction(UnicodeEmoji('🤔'));
+    }
+  });
+
+  // This event is called when a reaction has been added to a message
+  bot.eventsWs.onMessageReactionAdded.listen((event) {
+    final emoji = (event.emoji as IGuildEmojiPartial).isPartial ? (event.emoji as IGuildEmojiPartial).resolve() : event.emoji;
+    if(emoji is UnicodeEmoji) {
+      event.message?.channel?.sendMessage(MessageBuilder.content('Woah! This is a unicode emoji: ${event.emoji}'));
```
abitofevrything, line 28: The null-aware accessor on `channel` isn't necessary as it isn't nullable.

example/emojis.dart
```diff
@@ -0,0 +1,34 @@
...
+      final msg = await event.message.channel.sendMessage(MessageBuilder.content('Look at this emoji: $emoji'));
+      msg.createReaction(emoji);
+      // For unicode emoji use `UnicodeEmoji` class
+      msg.createReaction(UnicodeEmoji('🤔'));
+    }
+  });
+
+  // This event is called when a reaction has been added to a message
+  bot.eventsWs.onMessageReactionAdded.listen((event) {
+    final emoji = (event.emoji as IGuildEmojiPartial).isPartial ? (event.emoji as IGuildEmojiPartial).resolve() : event.emoji;
```
abitofevrything, line 26: This will throw a cast error if the emoji is a unicode emoji.

lib/src/core/message/guild_emoji.dart
```diff
@@ -14,41 +15,87 @@ abstract class IBaseGuildEmoji implements SnowflakeEntity, IEmoji {
...
   String get cdnUrl;
 
+  /// The name of the emoji.
+  String get name;
+
+  /// Whether this emoji is animated.
+  bool get animated;
+
   /// Creates partial emoji from given String or Snowflake.
-  factory IBaseGuildEmoji.fromId(Snowflake id) => GuildEmojiPartial(id);
```
abitofevrything, line 18: To avoid breaking changes, could you keep the `fromId` constructor and have it return a partial emoji that can be resolved with the client's emoji cache?

lib/src/core/message/reaction.dart
```diff
@@ -15,6 +21,10 @@ abstract class IReaction {
 
 /// Reaction object. [emoji] field can be partial [GuildEmoji].
 class Reaction implements IReaction {
+  /// Reference to [NyxxWebsocket]
+  @override
+  late final INyxx client;
```
abitofevrything, line 26: The `late` here isn't necessary.

abitofevrything (changes requested):
Looks great, apart from a couple of quick issues. Thank you so much for this contribution!

lib/src/core/message/guild_emoji.dart
Rapougnac: Yup, fixed it, sorry for this mistake

example/emojis.dart
Rapougnac: Fixed

example/emojis.dart
Rapougnac: Yeah, fixed

example/emojis.dart
Rapougnac: True, fixed

example/emojis.dart
Rapougnac: That's right, fixed it

lib/src/core/message/guild_emoji.dart
Rapougnac: Yeah, sorry, I didn't think about it at all

lib/src/core/message/reaction.dart
Rapougnac: It is, otherwise dart will complain that client has not been initialized.
That's due to the [`Reaction.event()`](https://github.com/nyxx-discord/nyxx/blob/68ebcde5c7fdf37a5dacb71777c02de988ad9afc/lib/src/core/message/reaction.dart#L45) method below.

lib/src/core/message/guild_emoji.dart
Rapougnac: Fixed!

lib/src/core/message/reaction.dart
abitofevrything: Ah, I missed that. Thanks for pointing that out!

abitofevrything (approved):

abitofevrything (approved):

lib/src/core/message/guild_emoji.dart
```diff
@@ -70,6 +117,12 @@ abstract class IGuildEmoji implements IBaseGuildEmoji {
   /// whether this emoji is animated
   bool get animated;
 
+  /// The user that created this emoji, not present if [fetchCreator] was not called
+  Cacheable<Snowflake, IUser>? get creator;
```
l7ssha, line 121: This api feels weird, why not provide single function for featching creator without that kind of weird state?

lib/src/core/guild/guild.dart
```diff
@@ -520,6 +520,9 @@ class Guild extends SnowflakeEntity implements IGuild {
       raw["emojis"].forEach((dynamic o) {
         final emoji = GuildEmoji(client, o as RawApiMap, id);
         emojis[emoji.id] = emoji;
+        if (client.cacheOptions.emojiCachePolicy.canCache(emoji) && client.cacheOptions.emojiCachePolicyLocation.http) {
+          client.emojis[emoji.id] = emoji;
+        }
```
l7ssha, line 525: why we need global cache for emojis?

l7ssha: Even if we need that kind of cache it would be better to interate over guilds are flatmap all emojis from guilds instead creating another cache storage

l7ssha: Also emoji cache policies and thingies are not applied to caching emojis inside guild object which is non consistent with rest of your implementation and whole lib

l7ssha (changes requested):

lib/src/core/message/guild_emoji.dart
```diff
@@ -14,41 +15,87 @@ abstract class IBaseGuildEmoji implements SnowflakeEntity, IEmoji {
...
-  GuildEmojiPartial(Snowflake id) : super({"id": id.toString()});
+  GuildEmojiPartial(RawApiMap raw, [this.client]) : super({"id": raw["id"].toString()}) {
+    name = raw["name"] as String? ?? "nyxx";
+    animated = raw["animated"] as bool? ?? false;
+  }
+
+  /// Resolves this [GuildEmojiPartial] to [GuildEmoji]
+  @override
+  IGuildEmoji? resolve() {
+    if (client != null) {
```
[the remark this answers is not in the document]
Rapougnac: I agree, I'll try to fix this asap

lib/src/core/guild/guild.dart
Rapougnac: > why we need global cache for emojis?

I thought it would be useful to access directly by the client

lib/src/core/message/guild_emoji.dart
l7ssha: yeah sure. Then it would be better since caching cacheable in emojis is not neccessary

lib/src/core/message/guild_emoji.dart
```diff
@@ -102,19 +155,43 @@ class GuildEmoji extends BaseGuildEmoji implements IGuildEmoji {
...
+    name = raw["name"] as String;
     requireColons = raw["require_colons"] as bool? ?? false;
     managed = raw["managed"] as bool? ?? false;
     animated = raw["animated"] as bool? ?? false;
     roles = [for (final roleId in raw["roles"]) RoleCacheable(client, Snowflake(roleId), guild)];
   }
 
+  /// Fetches the creator of this emoji
+  @override
+  Future<IUser> fetchCreator() => client.httpEndpoints.fetchEmojiCreator(guild.id, id);
```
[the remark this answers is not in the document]
l7ssha: sure thing. Part of comments were add before getting into latter parts of code so I didn't had full undrestanding of your implementation. That methods should be left here and user property from emoji removed as in my other comment :)

lib/src/core/guild/guild.dart
Rapougnac: Yeah right, so I'll remove this property

lib/src/core/message/guild_emoji.dart
```diff
@@ -14,41 +15,121 @@ abstract class IBaseGuildEmoji implements SnowflakeEntity, IEmoji {
...
   /// Creates an instance of [GuildEmojiPartial]
-  GuildEmojiPartial(Snowflake id) : super({"id": id.toString()});
+  GuildEmojiPartial(RawApiMap raw, [this.client]) : super({"id": raw["id"].toString()}) {
+    name = raw["name"] as String? ?? "nyxx";
+    animated = raw["animated"] as bool? ?? false;
+  }
+}
+
+
+
```
l7ssha, line 102: there should be only one line

lib/src/core/message/guild_emoji.dart
```diff
@@ -14,41 +15,121 @@ abstract class IBaseGuildEmoji implements SnowflakeEntity, IEmoji {
...
+abstract class IResolvableGuildEmojiPartial implements IGuildEmojiPartial {
+  /// Resolves this [IResolvableGuildEmojiPartial] to [IGuildEmoji]
+  IGuildEmoji resolve();
+}
 
 class GuildEmojiPartial extends BaseGuildEmoji implements IGuildEmojiPartial {
+  /// Reference to [INyxxWebsocket]
+  @override
+  INyxx? client;
```
l7ssha, line 76: its not needed here

lib/src/core/message/guild_emoji.dart
```diff
@@ -14,41 +15,121 @@ abstract class IBaseGuildEmoji implements SnowflakeEntity, IEmoji {
...
+
+  /// Whether this emoji can be resolved to a [IGuildEmoji]
+  @override
+  bool get isResolvable => this is IResolvableGuildEmojiPartial && client != null;
```
l7ssha, line 80: that property is useless since end developer can check if objecys instance is resolvable.

lib/src/core/message/guild_emoji.dart
```diff
@@ -14,41 +15,121 @@ abstract class IBaseGuildEmoji implements SnowflakeEntity, IEmoji {
...
+}
+
+
+
+class ResolvableGuildEmojiPartial extends BaseGuildEmoji implements IResolvableGuildEmojiPartial {
+  /// Whether this emoji is animated.
+  @override
+  late final bool animated;
+
+  /// Reference to [INyxxWebsocket]
```
l7ssha, line 108: functionality does not require websocket connection so this doc is wrong

lib/src/core/message/reaction.dart
```diff
@@ -28,21 +38,20 @@ class Reaction implements IReaction {
   late final IEmoji emoji;
 
   /// Creates an instance of [Reaction]
-  Reaction(RawApiMap raw) {
+  Reaction(RawApiMap raw, this.client) {
```
l7ssha, line 41: if client is passed only for emoji in constructor then class property its not needed

lib/src/internal/http_endpoints.dart
```diff
@@ -556,6 +559,25 @@ class HttpEndpoints implements IHttpEndpoints {
     return Future.error(response);
   }
 
+  @override
+  Future<IUser> fetchEmojiCreator(Snowflake guildId, Snowflake emojiId) async {
+    final response = await httpHandler.execute(BasicRequest("/guilds/$guildId/emojis/$emojiId"));
+
+    if (response is HttpResponseSuccess) {
+      if(response.jsonBody["managed"] as bool) {
+        return Future.error("Emoji is managed");
```
l7ssha, line 568: can you wrap returned error string with ArgumentError class. Thats the pattern in whole file. Its not ideal, but Im gonna deal with it in the future

l7ssha (changes requested):

lib/src/core/message/guild_emoji.dart
```diff
@@ -14,41 +15,121 @@ abstract class IBaseGuildEmoji implements SnowflakeEntity, IEmoji {
...
+  @override
+  bool get isPartial => true;
+
+  /// The name of the emoji.
+  @override
+  late final String name;
+
+  /// Resolves this [IResolvableGuildEmojiPartial] to [IGuildEmoji]
+  @override
+  IGuildEmoji resolve() => client.guilds.values.expand((guild) => guild.emojis.values).toList().firstWhere((emoji) => emoji.id == id) as IGuildEmoji;
```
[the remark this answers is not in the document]
Rapougnac: Nope, I added it but I don't know why

lib/src/core/message/reaction.dart
Rapougnac: But, there's the `Reaction.event` method below

l7ssha: I dont understand what you mean. Public property of INyxx in Reaction is not needed since its used only to pass to  `ResolvableGuildEmojiPartial` in line 49. you can make simple parameter and pass it there

lib/src/core/message/guild_emoji.dart
```diff
@@ -12,29 +14,43 @@ abstract class IBaseGuildEmoji implements SnowflakeEntity, IEmoji {
   bool get isPartial;
 
   /// Returns cdn url to emoji
-  String get cdnUrl;
+  String cdnUrl({String? format = "webp", int? size = 128});
```
l7ssha, line 17: method should be below constructors

lib/src/core/message/guild_emoji.dart
```diff
@@ -12,29 +14,43 @@ abstract class IBaseGuildEmoji implements SnowflakeEntity, IEmoji {
...
   bool get isPartial;
 
+  /// Whether this emoji is animated.
+  @override
+  bool get animated;
+
   /// Returns cdn url to emoji
   @override
-  String get cdnUrl => "https://cdn.discordapp.com/emojis/$id.png";
+  String cdnUrl({String? format = "webp", int? size = 128}) => "${Constants.cdnUrl}/emojis/$id.$format?size=$size";
```
l7ssha, line 40: format shouldn;t be null because if someone passes here null it would evaulable to 'null' in resulting string which is garbage.

Instead would be better in that case I think to leave both as nullable without value and if null provided for format then you automatically should check if emoji is animated or not and if user provided non null format then it should be used that. and whole `size` get param should be any appended to string if `size` param is not null

Something like:
```dart
String cdnUrl({String? format = "webp", int? size = 128}) {
  var url = "${Constants.cdnUrl}/emojis/$id.";

  if (format == null) {
    if (animated) {
      url += "gif";
    } else {
      url += "webp";
    }
  } else {
    url += format;
  }

  if (size != null) {
    url += "?size=$size";  
  }

  return url;
}
```

l7ssha (changes requested):

lib/src/core/message/reaction.dart
Rapougnac: Ohh, ok, got that, sorry, I didn't understood at first

lib/src/core/message/guild_emoji.dart
Rapougnac: I agree with that, but, shouldn't we let the developer choose if an animated emoji will be displayed as a static/dynamic image?

l7ssha: yeah, my example includes that case

Rapougnac: Like:
```dart
String cdnUrl({String? format = "webp", int? size = 128, bool? animate = false}) {
	var url = "${Constants.cdnUrl}/emojis/$id.";
	
	if (format == null) {
		if(animated && animate != null && animate) {
			url += "gif";
		} else {
			url += "webp";
		}
	} else {
		if (animated && animate != null && animate) {
			url += "gif";
		} else {
			url += format;
		}
	}

	if (size != null) {
		url += "?size=$size";
	}
	
	return url;
}
```

l7ssha: why do we need `animated` checks? If developer wants static emoji even if emoji is animated we should give him non animated url. So my version stands up:
```dart
String cdnUrl({String? format, int? size}) {
  var url = "${Constants.cdnUrl}/emojis/$id.";

  if (format == null) {
    if (animated) {
      url += "gif";
    } else {
      url += "webp";
    }
  } else {
    url += format;
  }

  if (size != null) {
    url += "?size=$size";  
  }

  return url;
}
```

lib/src/core/message/guild_emoji.dart
HarryET: +1 that

HarryET (changes requested):

lib/src/core/message/guild_emoji.dart
Rapougnac: Oh, that's right, didn't pay attention

HarryET (dismissed):
Looks good to merge 🎉

lib/src/core/message/guild_emoji.dart
```diff
@@ -1,40 +1,66 @@
...
-  String get cdnUrl;
+  /// The name of the emoji.
+  String get name;
+
+  /// Whether this emoji is animated.
+  bool get animated;
 
   /// Creates partial emoji from given String or Snowflake.
-  factory IBaseGuildEmoji.fromId(Snowflake id) => GuildEmojiPartial(id);
+  factory IBaseGuildEmoji.fromId(Snowflake id) => GuildEmojiPartial({"id": id});
```
l7ssha, line 15: Can it be returned as string? Because its inconsistent with BaseGuildEmoji since it takes snowflake as string

lib/src/core/message/guild_emoji.dart
```diff
@@ -1,40 +1,66 @@
...
+  /// The name of the emoji.
+  @override
+  String get name;
 
   /// Creates an instance of [BaseGuildEmoji]
   BaseGuildEmoji(RawApiMap raw) : super(Snowflake(raw["id"]));
 
+  /// Returns cdn url to emoji
+  @override
+  String cdnUrl({String? format = "webp", int? size = 128}) {
```
l7ssha, line 39:
```suggestion
  String cdnUrl({String? format, int? size}) {
```
It needs tp be null by deafault because otherwise it wont do thing automatically

l7ssha (changes requested):

l7ssha (approved):

l7ssha:
Looks good. Gonna merge today in the afternoon. Thanks for contributing :)

PR #307, opened 2022-03-05
Fix target id property and add guild audit logs options

Rapougnac:
Add options for guild audit logs, before was set as a string nullable
Also add new audit logs events & fix targetId.

## Type of change

	- `targetId` does not throw error anymore
	- Audit logs options is now `IAuditLogOptions?` instead of `String?`
# Checklist:

### DIFF
--- a/lib/src/core/audit_logs/audit_log_options.dart
+++ b/lib/src/core/audit_logs/audit_log_options.dart
@@ -1,0 +1,107 @@
+import 'package:nyxx/nyxx.dart';
+
+/// Additionnal info for certain [action types](https://discord.com/developers/docs/resources/audit-log#audit-log-entry-object-audit-log-events)
+abstract class IAuditLogOptions {
+  /// The channel in which the entites were targeted.
+  /// 
+  /// Shows only in: 
+  /// - [AuditLogEntryType.memberMove]; 
+  /// - [AuditLogEntryType.messagePin];
+  /// - [AuditLogEntryType.messageUnpin];
+  /// - [AuditLogEntryType.messageDelete];
+  /// - [AuditLogEntryType.stageInstanceCreate];
+  /// - [AuditLogEntryType.stageInstanceUpdate] and [AuditLogEntryType.stageInstanceDelete] action types
+  Snowflake? get channelId;
+
+  /// The number of entities targeted.
+  /// Shows only in: 
+  /// - [AuditLogEntryType.messageDelete];
+  /// - [AuditLogEntryType.messageBulkDelete];
+  /// - [AuditLogEntryType.memberDisconnect] and [AuditLogEntryType.memberMove] action types 
+  int? get count;
+
+  /// The number of days after which inactive users will be kicked.
+  /// Shows only in: [AuditLogEntryType.memberPrune] action types
+  String? get deleteMemberDelay;
+
+  /// Id of the overwritten entity.
+  /// Shows only in: 
+  /// - [AuditLogEntryType.channelOverwriteCreate];
+  /// - [AuditLogEntryType.channelOverwriteUpdate] and [AuditLogEntryType.channelOverwriteDelete] action types
+  Snowflake? get id;
+
+  /// The number of the members removed by the prune.
+  /// Shows only in: [AuditLogEntryType.memberPrune] action types
+  int? get pruneCount;
+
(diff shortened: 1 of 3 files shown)


abitofevrything:
This seems to be breaking due to the change of the `options` field type (`String?` -> `IAuditLogOptions?`), could you update the description accordingly?

Also, since this is a breaking change, it should target the `next` branch and not the `dev` branch. You can change the base branch by clicking the "Edit" button next to the title.

Rapougnac:
Sure ! Thought it was a bug fix because it was throwing a cast error with options

l7ssha:
Can you rebase or merge next into your branch. Commit history is unclear so is diff for review

lib/src/core/audit_logs/audit_log_options.dart
```diff
@@ -0,0 +1,93 @@
...
+  /// Type of overwritten entity.
+  /// One of:
+  ///  - `role`
+  ///  - `member`
+  @override
+  late final String? overwrittenType;
+
+  AuditLogOptions(RawApiMap raw) {
+    channelId = (raw['channel_id'] as String?)?.toSnowflake();
+    count = raw['count'] != null ? int.parse(raw['count'] as String) : null;
```
l7ssha, line 76: int parsing is probably redundant because if property in json is number its goint to be deserialized into number

lib/src/core/audit_logs/audit_log_options.dart
```diff
@@ -0,0 +1,93 @@
...
+    deleteMemberDuration = (raw['delete_member_days'] as String?) != null ? Duration(days: int.parse(raw['delete_member_days'] as String)) : null;
+    id = (raw['id'] as String?)?.toSnowflake();
+    pruneCount = (raw['members_removed'] as String?) != null ? int.parse(raw['members_removed'] as String) : null;
```
l7ssha, line 79: int parsing is probably redundant because if property in json is number its goint to be deserialized into number

lib/src/core/audit_logs/audit_log_options.dart
```diff
@@ -0,0 +1,93 @@
...
+  String? get roleName;
+
+  /// Type of overwritten entity.
+  /// One of:
+  ///   - `role`
+  ///   - `member`
+  String? get overwrittenType;
+
+  /// The constructor for the [IAuditLogOptions]
+  IAuditLogOptions(RawApiMap raw);
```
l7ssha, line 35: thats not needed here

lib/src/core/audit_logs/audit_log_options.dart
```diff
@@ -0,0 +1,93 @@
...
+
+  AuditLogOptions(RawApiMap raw) {
+    channelId = (raw['channel_id'] as String?)?.toSnowflake();
+    count = raw['count'] != null ? int.parse(raw['count'] as String) : null;
+    deleteMemberDuration = (raw['delete_member_days'] as String?) != null ? Duration(days: int.parse(raw['delete_member_days'] as String)) : null;
+    id = (raw['id'] as String?)?.toSnowflake();
+    pruneCount = (raw['members_removed'] as String?) != null ? int.parse(raw['members_removed'] as String) : null;
+    messageId = (raw['message_id'] as String?)?.toSnowflake();
+    roleName = raw['role_name'] as String?;
+    switch (raw['type']) {
```
l7ssha, line 82: here again if prop is number in json its going to be parsed as json so switch case wont work

l7ssha (changes requested):

lib/src/core/audit_logs/audit_log_options.dart
abitofevrything: This would actually fail because of casting an int to a String if that were the case.

Rapougnac: It's a string according to the docs: <https://discord.com/developers/docs/resources/audit-log#audit-log-entry-object-optional-audit-entry-info>

l7ssha: if that's string returned from api then its fine

l7ssha (approved):
Looks good. Thanks for contribution :)

l7ssha:
@Rapougnac can you merge next into your branch and resolve conflicts?

PR #313, opened 2022-03-14
Allow editing messages to remove content

### LINKED ISSUE #312: Allow empty message content when editing message
**Describe the bug**
When editing an message, users should be able to clear the content of the message *if* the message has an embed or attachment. We believe this is done by sending an empty string in the Update Message payload, but currently `MessageBuilder` does not include `content` in the payload if it is empty.

**To Reproduce**
Send a message with an embed, attempt to edit it to remove content.

**Expected behavior**
The message content should be removed.

 - OS: *
 - Dart version: *
 - Nyxx version: 3.3.0

abitofevrything:
# Description

Includes `content` field in message payload even if it is empty to allow users to edit their messages to remove content. Also includes some fixes concerning `late final` fields that were not always initialised correctly.

Closes #312

## Type of change

# Checklist:

### DIFF
--- a/lib/src/core/user/member.dart
+++ b/lib/src/core/user/member.dart
@@ -192,11 +192,6 @@ class Member extends SnowflakeEntity implements IMember {
 
     roles = [for (var id in raw["roles"]) RoleCacheable(client, Snowflake(id), guild)];
 
-    // TODO: remove this; hoisted_role is never present in member object
-    // if (raw["hoisted_role"] != null) {
-    //   hoistedRole = RoleCacheable(client, Snowflake(raw["hoisted_role"]), guild);
-    // }
-
     joinedAt = DateTime.parse(raw["joined_at"] as String).toUtc();
 
     if (client.cacheOptions.userCachePolicyLocation.objectConstructor) {
--- a/lib/src/core/user/member.dart
+++ b/lib/src/core/user/member.dart
@@ -131,6 +131,7 @@ class Member extends SnowflakeEntity implements IMember {
 
   /// Highest role of member
   @override
+  @Deprecated('Use `roles` and sort by position instead. Attempting to use this field will throw an error.')
   late final Cacheable<Snowflake, IRole> hoistedRole;
 
   /// When the user starting boosting the guild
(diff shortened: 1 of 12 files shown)


lib/src/core/user/member.dart
```diff
@@ -191,13 +192,12 @@ class Member extends SnowflakeEntity implements IMember {
 
     roles = [for (var id in raw["roles"]) RoleCacheable(client, Snowflake(id), guild)];
 
-    if (raw["hoisted_role"] != null) {
-      hoistedRole = RoleCacheable(client, Snowflake(raw["hoisted_role"]), guild);
-    }
+    // TODO: remove this; hoisted_role is never present in member object
+    // if (raw["hoisted_role"] != null) {
+    //   hoistedRole = RoleCacheable(client, Snowflake(raw["hoisted_role"]), guild);
+    // }
```
l7ssha, line 198: you can remove it as well, we dont need commented code

lib/src/internal/http_endpoints.dart
```diff
@@ -1125,6 +1125,10 @@ class HttpEndpoints implements IHttpEndpoints {
 
   @override
   Future<IMessage> editMessage(Snowflake channelId, Snowflake messageId, MessageBuilder builder) async {
+    if (!builder.canBeUsedAsNewMessage()) {
+      return Future.error(ArgumentError("Cannot edit a message to have neither content nor embeds"));
+    }
```
l7ssha, line 1130: is that intended here? Checks says `canBeUsedAsNewMessage` and we are editing here

l7ssha (changes requested):

lib/src/internal/http_endpoints.dart
abitofevrything: Yes. The resulting message from the edit has to be a valid message, so `canBeUsedAsNewMessage` is checked.

It isn't perfect as some edge cases are not caught but it is the cleanest non-breaking solution, as other solutions would need the builder to keep some reference to the original message.

l7ssha (approved):

PR #154, opened 2021-07-19
[feature] User banners

l7ssha:
# Description

This PR implements user banners and refactor some of color stuff

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# Checklist:

- [ ] Ran `dartanalyzer --options analysis_options.yaml .`
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #318, opened 2022-04-02
Add `@bannerUrl()` method

Rapougnac:
# Description
Add a `@bannerUrl` method on `IUser` class because all properties related to users banners was `bannerHash`.

Complete #154 

## Type of change
	- Add `@bannerUrl` method on `IUser`

# Checklist:

### DIFF
--- a/lib/src/internal/http_endpoints.dart
+++ b/lib/src/internal/http_endpoints.dart
@@ -170,6 +170,9 @@ abstract class IHttpEndpoints {
   /// Returns url to user avatar
   String userAvatarURL(Snowflake userId, String? avatarHash, int discriminator, {String format = "webp", int size = 128});
 
+  /// Returns url to user banner
+  String? userBannerURL(Snowflake userId, String? bannerHash, {String? format, int? size});
+
   /// Returns url to member avatar url
   String memberAvatarURL(Snowflake memberId, Snowflake guildId, String avatarHash, {String format = "webp"});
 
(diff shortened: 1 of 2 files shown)


lib/src/internal/http_endpoints.dart
```diff
@@ -1583,7 +1583,29 @@ class HttpEndpoints implements IHttpEndpoints {
       "${Constants.cdnUrl}/guilds/$guildId/users/$memberId/avatars/$avatarHash.$format";
 
   @override
-  String getUserBannerURL(Snowflake userId, String hash, {String format = "png"}) => "${Constants.cdnUrl}/banners/$userId/$hash.$format";
+  String? getUserBannerURL(Snowflake userId, String? hash, {String? format, int? size}) {
```
l7ssha, line 1586:
```suggestion
  String? userBannerURL(Snowflake userId, String? hash, {String? format, int? size}) {
```

lib/src/internal/http_endpoints.dart
```diff
@@ -1583,7 +1583,29 @@ class HttpEndpoints implements IHttpEndpoints {
...
+    if (hash != null) {
```
l7ssha, line 1587: it could be inverted so if hash is null you return null and otherwise rest of logic. it would removed one level of nesting

l7ssha (changes requested):

l7ssha (approved):

PR #346, opened 2022-05-27
Add missing emoji endpoints

### LINKED ISSUE #345: Implement all reaction endpoints
Some of the endpoints related to reactions, like [Get Reactions](https://discord.com/developers/docs/resources/channel#get-reactions) and [Delete All Reactions for Emoji](https://discord.com/developers/docs/resources/channel#delete-all-reactions-for-emoji), do not seem to be implemented. This makes features that rely on reaction interactions, like a *starboard*, difficult to materialize, where one has to keep track of which users reacted with what emoji to a given message.

In my opinion, [Get Reactions](https://discord.com/developers/docs/resources/channel#get-reactions) should be available on `IReaction` in the form of a property or an asynchronous method returning `List<Cacheable<Snowflake, IUser>>`/`List<IUser>`, or something similar. Currently, the `count` property of `int` type is all one can get.

As for [Delete All Reactions for Emoji](https://discord.com/developers/docs/resources/channel#delete-all-reactions-for-emoji), I'd propose a new method on `IMessage`, something like `Future<void> deleteReactions(IEmoji emoji)`.

abitofevrything:
Closes #345

### DIFF
--- a/lib/src/core/message/message.dart
+++ b/lib/src/core/message/message.dart
@@ -134,6 +134,12 @@ abstract class IMessage implements SnowflakeEntity, Disposable, Convertable<Mess
   /// Deletes all reactions
   Future<void> deleteAllReactions();
 
+  /// Fetches the users that reacted to this message with a given emoji.
+  Future<List<IUser>> fetchReactions(IEmoji emoji);
+
+  /// Deletes reactions to this message with a given emoji
+  Future<void> deleteReactions(IEmoji emoji);
+
   /// Deletes the message.
   Future<void> delete({String? auditReason});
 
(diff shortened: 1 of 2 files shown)


lib/src/core/message/message.dart
```diff
@@ -134,6 +134,12 @@ abstract class IMessage implements SnowflakeEntity, Disposable, Convertable<Mess
   /// Deletes all reactions
   Future<void> deleteAllReactions();
 
+  /// Fetches the users that reacted to this message with a given emoji.
+  Future<List<IUser>> fetchReactions(IEmoji emoji);
```
l7ssha, line 138: name is confusing - `fetchReactedUsersForEmoji` maybe ?

l7ssha: Also return type should be `Stream<IUser>`

lib/src/internal/http_endpoints.dart
```diff
@@ -1463,6 +1475,47 @@ class HttpEndpoints implements IHttpEndpoints {
         ..reactions(),
       method: "DELETE"));
 
+  @override
+  Future<List<IUser>> fetchMessageReactions(
```
l7ssha, line 1479: return type should be `Stream<IUser>`

l7ssha (changes requested):

l7ssha (dismissed):

l7ssha (approved):

Issue #350, opened 2022-06-03
bug: GuildEventBuilder serialization


Rapougnac:
**Describe the bug**
-
When building an Event with the `type` property, this serialized wrong, for example:
```dart
final eb = GuildEventBuilder()..type = GuildEventType.external;
print(eb.build());
// Gives:
// {
//   'channel_id': '0',
//   'entity_type': GuildEventType:3
// }
```
**To Reproduce**
Given sample up

**Expected behavior**
`'entity_type'` should be `3` or `'3'`. 

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Linux
 - Dart version 2.17.3
 - Nyxx version 4.0.0-dev.1

**Additional context**

PR #351, opened 2022-06-09
Fix some serialization bugs

### LINKED ISSUE #349: bug: `AttachmentBuilder#getBase64()` wrong mime type
**Describe the bug**
When invoking an instance of `AttachmentBuilder` (all constructors) and call `getBase64()`, it returns a wrong mime type.

**To Reproduce**
Create an instance of `AttachmentBuilder`, call `getBase64`

**Expected behavior**
The mime type should be: `data:image/{extension};base64,somelongencodeddata`, and currently, this is `data:image/.{extension};base64,somelongencodeddata`

**Screenshots**
None

 - OS: Linux
 - Dart version 2.17.3
 - Nyxx version 4.0.0-dev.1

**Additional context**
None

Rapougnac:
# Description

Fix some serialization problems, `getBase64()` was working tho, but from discord's docs;

Image data is a [Data URI scheme](https://en.wikipedia.org/wiki/Data_URI_scheme) that supports JPG, GIF, and PNG formats. An example Data URI format is:
```
data:image/jpeg;base64,BASE64_ENCODED_JPEG_IMAGE_DATA
```

Fixes #349 and #350

No

## Type of change

# Checklist:

### DIFF
--- a/lib/src/utils/builders/attachment_builder.dart
+++ b/lib/src/utils/builders/attachment_builder.dart
@@ -60,8 +60,7 @@ class AttachmentBuilder {
 
   /// Returns attachment encoded in Data URI scheme format
   /// See: https://discord.com/developers/docs/reference#image-data
-  String getBase64({String? defaultFormat = 'png'}) {
-    defaultFormat ??= 'png';
+  String getBase64({String defaultFormat = 'png'}) {
     final encodedData = base64Encode(_bytes);
     final fileExtension = path_utils.extension(_name);
     final extension = fileExtension.isNotEmpty ? fileExtension.substring(1) : defaultFormat;
--- a/lib/src/utils/builders/attachment_builder.dart
+++ b/lib/src/utils/builders/attachment_builder.dart
@@ -60,9 +60,11 @@ class AttachmentBuilder {
 
   /// Returns attachment encoded in Data URI scheme format
   /// See: https://discord.com/developers/docs/reference#image-data
-  String getBase64() {
+  String getBase64({String? defaultFormat = 'png'}) {
+    defaultFormat ??= 'png';
     final encodedData = base64Encode(_bytes);
-    final extension = path_utils.extension(_name).substring(1);
+    final fileExtension = path_utils.extension(_name);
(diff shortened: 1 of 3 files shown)


lib/src/utils/builders/attachment_builder.dart
```diff
@@ -62,7 +62,7 @@ class AttachmentBuilder {
   /// See: https://discord.com/developers/docs/reference#image-data
   String getBase64() {
     final encodedData = base64Encode(_bytes);
-    final extension = path_utils.extension(_name);
+    final extension = path_utils.extension(_name).substring(1);
```
abitofevrything, line 65: This will throw if the file has no exception. It might be a good idea to include a "sensible default" like PNG in case the image file has no extension

abitofevrything: Yeah - maybe extract `path_utils.extension(_name)` to a variable so we don't compute it twice but that's the idea. Even better - make `'png'` an optional named parameter named `defaultFormat`.

lib/src/utils/builders/attachment_builder.dart
```diff
@@ -60,9 +60,11 @@ class AttachmentBuilder {
 
   /// Returns attachment encoded in Data URI scheme format
   /// See: https://discord.com/developers/docs/reference#image-data
-  String getBase64() {
+  String getBase64({String? defaultFormat = 'png'}) {
+    defaultFormat ??= 'png';
```
abitofevrything, line 64: No need for this, the parameter can be non nullable with a default value

abitofevrything (changes requested):

lib/src/utils/builders/attachment_builder.dart
l7ssha: better to remove `?` from type. Makes no sense if default parameter is present

Rapougnac: Yeah, sorry, I'm not paying much attention today 😅

abitofevrything (approved):

PR #364, opened 2022-08-22
feature: Event to notify change of connection status; Fixes #363

### LINKED ISSUE #363: Event to notify change of connection status
**Is your feature request related to a problem? Please describe.**
I would like to know when the bot loses connection for any reason, and when it gets back online.

**Describe the solution you'd like**
An event to subscribe to.

**Describe alternatives you've considered**
Using this hacky code for now:
```dart
class HackyLogging extends BasePlugin {
  @override
  FutureOr<void> onRegister(INyxx nyxx, logg.Logger logger) {
    logg.Logger.root.onRecord.listen((logg.LogRecord rec) {
      //TODO: Need to keep an eye on 
      // https://github.com/nyxx-discord/nyxx/blob/dev/lib/src/internal/shard/shard.dart#L221
      // in case the string changes even slightly!
      if( rec.message.startsWith("Not received previous heartbeat ack") ){
        print("Oh no!!!");
      } else if( rec.message.contains("connected to gateway") ){
        // Catch up with all the stuff since we were last online...
      }
      print("[${rec.time}] [${rec.level.name}] [${rec.loggerName}] ${rec.message}");
    });
  }
}

.
.
.

bot!.registerPlugin(HackyLogging());
```

l7ssha:
# Description

Add hook to `BasePlugin` that allows listening to web socket connection changes

Fixes #363

## Type of change

# Checklist:

### DIFF
--- a/lib/src/internal/shard/shard.dart
+++ b/lib/src/internal/shard/shard.dart
@@ -1,6 +1,8 @@
 import 'dart:async';
+import 'dart:html';
 import 'dart:io';
 import 'dart:isolate';
+import 'dart:js';
 
 import 'package:nyxx/src/core/snowflake.dart';
 import 'package:nyxx/src/core/guild/client_user.dart';
@@ -233,10 +235,18 @@ class Shard implements IShard {
       return;
     }
 
+    final closeReason = data['errorReason'] as String?;
+    final socketError = data['error'] as String?;
+
+    for (final plugin in manager.connectionManager.client.plugins) {
+      plugin.onConnectionChange(manager.connectionManager.client, manager.logger, closeCode, closeReason, socketError);
+    }
+
     _connected = false;
     _heartbeatTimer.cancel();
-    manager.logger.severe("Shard $id disconnected. Error: [${data['error']}] Error code: [${data['errorCode']}] | Error message: [${data['errorReason']}]");
     manager.onDisconnectController.add(this);
+
+    manager.logger.severe("Shard $id disconnected. Error: [$socketError] Error code: [$closeCode] | Error message: [$closeReason]");
 
     switch (closeCode) {
       case 4004:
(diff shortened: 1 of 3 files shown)


lib/src/internal/shard/shard.dart
```diff
@@ -1,6 +1,8 @@
 import 'dart:async';
+import 'dart:html';
```
abitofevrything, line 2: Unneeded `dart:html` import will prevent compilation

lib/src/internal/shard/shard.dart
```diff
@@ -1,6 +1,8 @@
...
 import 'dart:io';
 import 'dart:isolate';
+import 'dart:js';
```
abitofevrything, line 5: Unneeded `dart:js` import will prevent compilation

lib/src/plugin/plugin.dart
```diff
@@ -8,4 +8,6 @@ abstract class BasePlugin {
 
   FutureOr<void> onBotStart(INyxx nyxx, Logger logger) async {}
   FutureOr<void> onBotStop(INyxx nyxx, Logger logger) async {}
+
+  FutureOr<void> onConnectionChange(INyxx nyxx, Logger logger, int? closeCode, String? closeReason, String? socketError) async {}
```
abitofevrything, line 12: Should this not be another event stream, like [onDisconnect](https://pub.dev/documentation/nyxx/latest/nyxx/IWebsocketEventController/onDisconnect.html)? This would allow users to subscribe to changes without needing to resort to a plugin.

Plugin users can still use the event by subscribing to it in `onBotStart` or `onRegister`.

abitofevrything (changes requested):

lib/src/plugin/plugin.dart
l7ssha: we have already events for disconnects or other stuff I dont think we need another stream for agregating unrelated events (in context to code) so this seems like perfect candidate for a plugin.

abitofevrything (approved):
LGTM, though I'm not sure how this does much else compared to onDisconnect.

PR #380, opened 2022-10-26
feat: Add support for new select menus components

Rapougnac:
# Description
Add support for new types of select menus

- https://github.com/nyxx-discord/nyxx_interactions/pull/62

## Type of change

# Checklist:

### DIFF
--- a/lib/src/core/message/components/message_component.dart
+++ b/lib/src/core/message/components/message_component.dart
@@ -110,7 +110,7 @@ abstract class MessageComponent implements IMessageComponent {
   ComponentType get type;
 
   @override
-  late final String customId; 
+  late final String customId;
 
   MessageComponent(RawApiMap raw) {
     customId = raw['custom_id'] as String;
--- a/lib/src/core/message/components/message_component.dart
+++ b/lib/src/core/message/components/message_component.dart
@@ -239,7 +239,7 @@ abstract class IMessageMultiselect implements MultiSelectAbstract {
 
 class MessageMultiselect extends MultiSelectAbstract implements IMessageMultiselect {
   @override
-  ComponentType get type => ComponentType.select;
+  ComponentType get type => ComponentType.multiSelect;
 
   /// Possible options of multiselect
   @override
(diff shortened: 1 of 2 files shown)


lib/src/core/message/components/message_component.dart
```diff
@@ -77,16 +94,22 @@ class MessageComponentEmoji implements IMessageComponentEmoji {
...
+  IMessageComponent._(RawApiMap raw) {
+    customId = raw['custom_id'] as String;
+  }
 }
 
 /// Generic container for components that can be attached to message
-abstract class MessageComponent implements IMessageComponent {
-  /// [ComponentType]
+abstract class MessageComponent extends IMessageComponent {
+  /// Type of component.
```
abitofevrything, line 108: Any implementation should go in the non-interface class (without an `I` prefix). Therefore all implementation classes should implement and not extend their interfaces, and any functionality should go in that class.

abitofevrything (changes requested):

lib/src/core/message/components/message_component.dart
Rapougnac: Oops, my bad, you're right, seems I've messed up my head writing too much interfaces..

abitofevrything (dismissed):

lib/src/core/message/components/message_component.dart
```diff
@@ -98,17 +126,43 @@ abstract class MessageComponent implements IMessageComponent {
...
+  late final int maxValues;
+
+  /// Whether this multiselect is disabled.
+  late final bool isDisabled;
+
+  MultiSelectAbstract(RawApiMap raw) : super(raw) {
+    placeholder = raw['placeholder'] as String?;
+    minValues = raw['min_values'] as int? ?? 1;
+    maxValues = raw['max_values'] as int? ?? 1;
+    isDisabled = raw['disabled'] as bool? ?? false;
```
l7ssha, line 160: are these defaults based on something or are arbitrary?

l7ssha (approved):

lib/src/core/message/components/message_component.dart
Rapougnac: They're based on Discord's behaviour, whereas by default, a select menu can take only 1 to 1 (min, max) values, and are enabled.

l7ssha: good

Issue #382, opened 2022-11-01
Add websocket reconnection backoff 


l7ssha:
Hey man, 

Please allow us to set the time in between retries.
Ideally, provide a way to set three parameters: min (const Duration), max (const Duration), and multiplier (double).
(assert that max is larger than min or if it's null, don't impose an upper limit. multiplier should be >= 1)

E.g. I'd like it to start at 15s, doubling after each failure up to 10min. (This is to save battery.)
min:15s, max:10min, multiplier:2 ===> 15s, 30s, 1min, 2min, 4min, 8min, 10min, 10min, 10min, 10min, ....

Thanks a lot!

_Originally posted by @r3a1d3a1 in https://github.com/nyxx-discord/nyxx/issues/306#issuecomment-1295186712_

r3a1d3a1:
And obviously, once the connections succeeds, the delay should be reset to _min_ for future retries.

r3a1d3a1:
I hope you don't mind me asking. How can I feed in a custom RetryOptions so that minDelay is 10s and maxDelay is 10min?

r3a1d3a1:
Sorry, I meant to ask where in the code setup I feed my custom RetryOptions? CacheOptions or ClientOptions?!

abitofevrything:
ClientOptions... Retrying has nothing to do with caching

r3a1d3a1:
Awesome... Had to look around a bit to figure out how/where to pass it... That was the tricky part. Thank you!

PR #395, opened 2022-11-13
Add retry with backoff to network operations (gateway and http)

### LINKED ISSUE #394: Add http request backoff
Add back off for http retrying http requests (ex. to avoid problems with missing internet)

Related to #382

abitofevrything:
# Description

Ensures that shard connections and http requests retry if they fail due to a native exception (e.g due to a temporary internet outage) and that they do so with exponential backoff.

Closes #394 and #382.

## Type of change

# Checklist:

### DIFF
--- a/lib/src/client_options.dart
+++ b/lib/src/client_options.dart
@@ -1,6 +1,7 @@
 import 'package:nyxx/nyxx.dart';
 import 'package:nyxx/src/nyxx.dart';
 import 'package:nyxx/src/internal/shard/shard.dart';
+import 'package:retry/retry.dart';
 
 /// Options for configuring cache. Allows to specify where and which entities should be cached and preserved in cache
 class CacheOptions {
@@ -69,19 +70,24 @@ class ClientOptions {
   /// Allows to enable receiving raw gateway event
   bool dispatchRawShardEvent;
 
+  /// The [RetryOptions] to use when a shard fails to connect to the gateway.
+  RetryOptions shardReconnectOptions;
+
   /// Makes a new `ClientOptions` object.
-  ClientOptions(
-      {this.allowedMentions,
-      this.shardCount,
-      this.messageCacheSize = 100,
-      this.largeThreshold = 50,
-      this.compressedGatewayPayloads = true,
-      this.guildSubscriptions = true,
-      this.initialPresence,
-      this.shutdownHook,
-      this.shutdownShardHook,
-      this.dispatchRawShardEvent = false,
-      this.shardIds});
+  ClientOptions({
+    this.allowedMentions,
+    this.shardCount,
+    this.messageCacheSize = 100,
+    this.largeThreshold = 50,
+    this.compressedGatewayPayloads = true,
+    this.guildSubscriptions = true,
+    this.initialPresence,
+    this.shutdownHook,
+    this.shutdownShardHook,
+    this.dispatchRawShardEvent = false,
+    this.shardIds,
+    this.shardReconnectOptions = const RetryOptions(maxAttempts: 50), // 50 attempts won't ever reasonably be reached
+  });
 }
 
 /// When identifying to the gateway, you can specify an intents parameter which
--- a/lib/src/client_options.dart
+++ b/lib/src/client_options.dart
@@ -73,6 +73,12 @@ class ClientOptions {
   /// The [RetryOptions] to use when a shard fails to connect to the gateway.
   RetryOptions shardReconnectOptions;
 
+  /// The [RetryOptions] to use when a HTTP request fails.
+  ///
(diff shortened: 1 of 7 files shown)


lib/src/client_options.dart
```diff
@@ -69,19 +70,31 @@ class ClientOptions {
...
+    this.largeThreshold = 50,
+    this.compressedGatewayPayloads = true,
+    this.guildSubscriptions = true,
+    this.initialPresence,
+    this.shutdownHook,
+    this.shutdownShardHook,
+    this.dispatchRawShardEvent = false,
+    this.shardIds,
+    this.shardReconnectOptions = const RetryOptions(maxAttempts: 50), // 50 attempts won't ever reasonably be reached
+    this.httpRetryOptions = const RetryOptions(maxAttempts: 50),
```
l7ssha, line 96: i think we should lower max attempts here. I think max of ~10 should be enough and other than that should require manual intervention

lib/src/internal/http/http_handler.dart
```diff
@@ -78,25 +78,14 @@ class HttpHandler {
...
-      _upsertBucket(request, response);
-
-      // Return http error
-      return _handle(request, response);
-    }
+    currentBucket?.addInFlightRequest(request);
+    final response = await client.options.httpRetryOptions.retry(
+      () async => httpClient.send(await request.prepareRequest()),
+      onRetry: (ex) => logger.shout('Exception when sending HTTP request (retrying automatically): $ex'),
+    );
```
l7ssha, line 85: shoudn't it be implemented similarly to package example, that we retry only on TimeoutException or similar?

l7ssha (approved):

lib/src/internal/http/http_handler.dart
abitofevrything: We should retry on any exception, which include `TImeoutException` but also other common exceptions like `SocketException`, most of which are wrapped in `ClientException` by the http package anyways.

abitofevrything: We should retry on any exception, which include `TImeoutException` but also other common exceptions like `SocketException`, most of which are wrapped in `ClientException` by the http package anyways.

l7ssha: what if our out data we send is wrong (server responds with 400)? It will fail anyway but this time ~10 times

abitofevrything: A 4xx response code is still considered a success by the http package. It's up to us to determine if the response is an error or not, which is handled in `HttpHandler._handle` (converting the response from the http package into an `HttpResponseSuccess` or `HttpResponseError` depending on it's status code) and in `HttpEndpoints.executeSafe`, which throes if the response is not an `HttpResponseSuccess`).

This function will only retry if the http package throws an exception, not if we throw one as that is handled elsewhere. So no, as long as the request is correctly sent and received (even if it represents an error) then this function will not retry.

See the docs added for `ClientOptions` which also mention this.

l7ssha: i think now we consider HttpResponseSuccess if status code is >= 200 < 300. I think we should extend this behavior to correctly handle connection problems/timeouts since retrying on 400 doesnt make sense (and now it will try to retry on 400 since we are throwing exception if response is not HttpResponseSuccess

abitofevrything: No - `HttpResponse` from nyxx isn't in the picture here. Only the `http` package, which only throws on actual errors and not responses with a 4xx response code.

Throwing `HttpResponseError`s is done elsewhere using the return value of this function. Since the place where that is handled is *not* wrapped in retry, that will not trigger a retry.

As long as the server responds, even with an erroneous code, an exception will not be thrown here and this will not retry,

l7ssha: yeah sure i do understand that. But we cant implement that feature without excluding HTTP 400 s for example because its gonna try send bad data for example 10 times. We have to do something with that. Either by wrapping http behavior and throwing expection or doing something else to go around that problem

l7ssha (approved):

lib/src/client_options.dart
```diff
@@ -69,19 +70,24 @@ class ClientOptions {
...
+    this.messageCacheSize = 100,
+    this.largeThreshold = 50,
+    this.compressedGatewayPayloads = true,
+    this.guildSubscriptions = true,
+    this.initialPresence,
+    this.shutdownHook,
+    this.shutdownShardHook,
+    this.dispatchRawShardEvent = false,
+    this.shardIds,
+    this.shardReconnectOptions = const RetryOptions(maxAttempts: 50), // 50 attempts won't ever reasonably be reached
```
r3a1d3a1: Based on [this](https://github.com/google/dart-neats/blob/master/retry/lib/retry.dart#L107), it actually never exceeds 31 :)

PR #399, opened 2022-11-15
hotfix: components deserialization

Rapougnac:
# Description
Fix a regression that was introduced in  #380 

## Type of change

# Checklist:

### DIFF
--- a/lib/src/core/message/components/message_component.dart
+++ b/lib/src/core/message/components/message_component.dart
@@ -99,7 +99,8 @@ abstract class IMessageComponent {
   /// The [ComponentType]
   ComponentType get type;
 
-  /// The custom id of this component, set by the user.
+  /// The custom id of this component, set by the user, if there isn't one, an empty string is returned.
+  // TODO: Add nullable string in next major release.
   String get customId;
 }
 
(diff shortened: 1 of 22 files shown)


abitofevrything (commented):
This would be a breaking change but it would fix an unavoidable issue - @l7ssha what do you think?

We could merge the breaking change to fix the issue, which shouldn't cause too many issues as this was only added recently. We could also default the `customId` to an empty string if it wasn't set, or provide some form of alternative method and deprecate the existing one...

l7ssha:
@Rapougnac since its hotfix, add changelog and change target to main

lib/src/core/message/components/message_component.dart
```diff
@@ -99,8 +99,8 @@ abstract class IMessageComponent {
   /// The [ComponentType]
   ComponentType get type;
 
-  /// The custom id of this component, set by the user.
-  String get customId;
+  /// The custom id of this component set by the user, if any.
+  String? get customId;
```
abitofevrything, line 103: Add a TODO to remind us to make it nullable in the next major release as well please

abitofevrything (changes requested):

PR #403, opened 2022-11-18
Improve error handling and logging

### LINKED ISSUE #401: feature: improve error handling
**Is your feature request related to a problem? Please describe.**
sometimes error messages returned by nyxx cane be unclear and ambiguous

**Describe the solution you'd like**
Since discord returns error message with code and if request had payload: detailed info about whats wrong  with request we should be able to implement better messages using this info.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

abitofevrything:
# Description

Improves error handling, adds guidelines for logging practices and updates logging to match those practices.

Closes #401.

## Type of change

# Checklist:

### DIFF
--- a/lib/src/internal/http/http_response.dart
+++ b/lib/src/internal/http/http_response.dart
@@ -2,6 +2,8 @@ import 'dart:convert';
 import 'dart:typed_data';
 
 import 'package:http/http.dart' as http;
+import 'package:nyxx/nyxx.dart';
+import 'package:nyxx/src/internal/response_wrapper/error_response_wrapper.dart';
 
 /// Represents a HTTP result from the API.
 abstract class IHttpResponse {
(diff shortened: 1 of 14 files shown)


lib/src/nyxx.dart
```diff
@@ -202,22 +200,30 @@ class NyxxRest extends INyxxRest {
 
   @override
   Future<void> connect({bool propagateReady = true}) async {
+    _logger.config([
+      'Token: $token',
```
Rapougnac, line 204: Is it a good idea to log the token to stdout?

abitofevrything: It's already logged in HTTP request logging, so it doesn't really matter. What would be a good idea is to add an option to `Logging` to censor the token in the output, so I'll do that.

Rapougnac (approved):

lib/src/internal/http/http_response.dart
```diff
@@ -1,107 +1,168 @@
...
+    bool hasJsonBody = false;
 
     try {
-      if (_body.isNotEmpty) {
-        _jsonBody = jsonDecode(utf8.decode(_body));
-      } else {
-        _jsonBody = null;
-      }
+      textBody = utf8.decode(body);
+      jsonBody = jsonDecode(textBody);
```
l7ssha, line 73: here we are storing "same" data twice. It is realistic that we are gonna access json data more than once? If not we could do getter and decode on the fly to presever some memory

l7ssha (approved):

lib/src/internal/http/http_response.dart
abitofevrything: No, but because HttpResponse objects are short lived anyways, we don't gain any memory benefit from storing it for a short amount of time against decoding on the fly.

In fact, HttpResponseError does access jsonBody more than once, so we save on the CPU cost of decoding it twice by storing it.

PR #420, opened 2023-01-11
feature: ETF Encoding

Rapougnac:
# Description
This adds a way to request etf-encoded payloads and unpack them to get proper payload.

> **Warning**
> Snowflakes are sent as int64 (or String, depends, love discord), and not String, like in json.
> This can be unstable, also

-/-

## Type of change

# Checklist:

### DIFF
--- a/lib/src/internal/shard/shard.dart
+++ b/lib/src/internal/shard/shard.dart
@@ -405,7 +405,7 @@ class Shard implements IShard {
         "intents": manager.connectionManager.client.intents,
         if (manager.connectionManager.client.options.initialPresence != null) "presence": manager.connectionManager.client.options.initialPresence!.build(),
         "shard": <int>[id, manager.totalNumShards],
-        if (manager.connectionManager.client.options.payloadEncoding == Encoding.json)
+        if (manager.connectionManager.client.options.payloadEncoding == Encoding.json && !manager.connectionManager.client.options.compressedGatewayPayloads)
           "compresstion": manager.connectionManager.client.options.payloadCompression,
       });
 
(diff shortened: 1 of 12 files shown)


pubspec.yaml
```diff
@@ -14,6 +14,7 @@ dependencies:
   logging: ^1.0.1
   path: ^1.8.0
   retry: ^3.1.0
+  eterl: ^0.0.2
```
abitofevrything, line 17: If eterl is not yet stable and its API might have breaking changes before 1.0.0, we should pin it's version instead of just requiring a minimum.

Rapougnac: The only breaking change that will come is with BigInts, as they're not well implemented. And discord does not send bigints.
But yeah, it's still good to pin.

lib/src/internal/shard/shard.dart
```diff
@@ -403,7 +404,9 @@ class Shard implements IShard {
         "large_threshold": manager.connectionManager.client.options.largeThreshold,
         "intents": manager.connectionManager.client.intents,
         if (manager.connectionManager.client.options.initialPresence != null) "presence": manager.connectionManager.client.options.initialPresence!.build(),
-        "shard": <int>[id, manager.totalNumShards]
+        "shard": <int>[id, manager.totalNumShards],
+        if (manager.connectionManager.client.options.payloadEncoding == Encoding.json && !manager.connectionManager.client.options.compressedGatewayPayloads)
+          "compresstion": manager.connectionManager.client.options.payloadCompression,
```
l7ssha, line 409:
```suggestion
          "compression": manager.connectionManager.client.options.payloadCompression,
```

l7ssha (changes requested):

lib/src/internal/shard/shard.dart
Rapougnac: Oops, my bad

l7ssha (approved):

PR #438, opened 2023-02-02
Fix update member roles equality

Rapougnac:
# Description
A comparison of list with `==` or `!=` will always result `false` or `true`. Because lists are never the same in dart (except from constants with canonicalisation). This also fixes the nickname not updating.

-/-

## Type of change

# Checklist:

### DIFF
--- a/lib/src/core/user/member.dart
+++ b/lib/src/core/user/member.dart
@@ -248,7 +248,7 @@ class Member extends SnowflakeEntity implements IMember {
       this.nickname = nickname;
     }
 
-    if (this.roles != roles) {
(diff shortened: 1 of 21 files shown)


lib/src/core/user/member.dart
```diff
@@ -248,7 +248,7 @@ class Member extends SnowflakeEntity implements IMember {
       this.nickname = nickname;
     }
 
-    if (this.roles != roles) {
+    if (!this.roles.every((role) => roles.contains(role.id))) {
```
l7ssha, line 251: i think thats only partial solution. what if user has role removed?

abitofevrything (approved):

Issue #426, opened 2023-01-18
feature: Implement "List Active Guild Threads"


phorcys420:
**Is your feature request related to a problem? Please describe.**
The "List Active Guild Threads" endpoint is not currently implemented.
https://discord.com/developers/docs/resources/guild#list-active-guild-threads

**Describe the solution you'd like**
``IGuild`` should have a method to list active guild threads.

CoocooFroggy:
I am getting:

```
[2023-02-16 21:27:11.183216] [SHOUT] [IgnoreExceptions] Unhandled exception was thrown
Error: type 'Null' is not a subtype of type 'bool' in type cast
Stack trace:
#0      new ThreadListResultWrapper (package:nyxx/src/internal/response_wrapper/thread_list_result_wrapper.dart:40:31)
#1      HttpEndpoints.fetchGuildActiveThreads (package:nyxx/src/internal/http_endpoints.dart:1642:12)
<asynchronous suspension>
```

I believe there is no `has_more` in the response, unlike
https://discord.com/developers/docs/resources/channel#list-public-archived-threads

PR #443, opened 2023-02-17
hasMore is optional for guild.fetchActiveThreads()

CoocooFroggy:
# Description

guild.fetchActiveThreads() does not have a `has_more` parameter in [the response](https://discord.com/developers/docs/resources/guild#list-active-guild-threads).

Fixes the issue mentioned here:
https://github.com/nyxx-discord/nyxx/issues/426#issuecomment-1434011164

## Type of change

# Checklist:

### DIFF
--- a/lib/src/internal/response_wrapper/thread_list_result_wrapper.dart
+++ b/lib/src/internal/response_wrapper/thread_list_result_wrapper.dart
@@ -11,7 +11,7 @@ abstract class IThreadListResultWrapper {
   List<IThreadMember> get selfThreadMembers;
 
   /// Whether there are potentially additional threads that could be returned on a subsequent call
-  bool? get hasMore;
+  bool get hasMore;
 }
 
 /// Wrapper of threads listing results.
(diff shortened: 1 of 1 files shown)


lib/src/internal/response_wrapper/thread_list_result_wrapper.dart
```diff
@@ -11,7 +11,7 @@ abstract class IThreadListResultWrapper {
   List<IThreadMember> get selfThreadMembers;
 
   /// Whether there are potentially additional threads that could be returned on a subsequent call
-  bool get hasMore;
+  bool? get hasMore;
```
Rapougnac, line 14: This is a breaking change

l7ssha: null in this case is equal to false so I think if discord returned null we can assume its false

CoocooFroggy: Done

l7ssha (approved):

PR #446, opened 2023-02-18
Release 4.5.0

l7ssha:


.gitignore
```diff
@@ -36,4 +36,3 @@ test-*.dart
 **/coverage/**
 coverage.json
 lcov.info
-.vscode/
```
abitofevrything, line 39: Don't see why we need to remove this. We don't have any nyxx-specific config, but developers (me included) store their launch configurations in here which shouldn't be committed.

README.md
```diff
@@ -83,4 +83,4 @@ Read [contributing document](https://github.com/nyxx-discord/nyxx/blob/dev/CONTR
 
 ## Credits 
 
- * [Hackzzila's](https://github.com/Hackzzila) for [nyx](https://github.com/Hackzzila/nyx).
+ * [Hackzzila's](https://github.com/Hackzzila) for [nyx](https://github.com/Hackzzila/nyx).
```
abitofevrything, line 86: Files should end in a blank line.

lib/src/internal/shard/shard_handler.dart
```diff
@@ -59,21 +63,29 @@ class ShardRunner implements Disposable {
...
   ///
   /// Calls jsonDecode and sends the data back to the manager.
-  void receive(String payload) => execute(ShardMessage(
-        ShardToManager.received,
-        seq: seq--,
-        data: jsonDecode(payload),
-      ));
+  void receive(/* List<int>|RawApiMap|String */ dynamic payload) {
+    if (payload is List<int>) {
+      payload = utf8.decode(zlib.decode(payload));
```
abitofevrything, line 68: Bit of a nit but we should either move the decompression somewhere else or update the doc comment to reflect that the data may still be compressed.

abitofevrything (changes requested):

lib/src/internal/shard/shard_handler.dart
l7ssha: That now for now I think since its PR for release. We can do bug fix later for that

.gitignore
Rapougnac: @abitofevrything Sorry, late reply, but .vscode is already ignored on top of the file

PR #448, opened 2023-02-21
documentation: Channel invites

CoocooFroggy:
# Description

Changes .createInvite documentation to be valid, and adds more descriptions to the options

# Checklist:

### DIFF
--- a/lib/src/core/channel/guild/guild_channel.dart
+++ b/lib/src/core/channel/guild/guild_channel.dart
@@ -32,7 +32,7 @@ abstract class IGuildChannel implements IMinimalGuildChannel {
   /// Returns effective permissions for [role] to this channel including channel overrides.
   Future<IPermissions> effectivePermissionForRole(IRole role);
 
-  /// Fetches and returns all channel"s [Invite]s
+  /// Fetches and returns all channel's [Invite]s
   ///
   /// ```
   /// var invites = await chan.fetchChannelInvites();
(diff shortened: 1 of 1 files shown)


lib/src/core/channel/guild/guild_channel.dart
```diff
@@ -157,10 +171,24 @@ abstract class GuildChannel extends MinimalGuildChannel implements IGuildChannel
...
+  /// [temporary] determines whether this invite only grants temporary
+  /// membership.
+  ///
+  /// If [unique] is true, Discord won't try to reuse a similar invite and will
+  /// instead always generate a new invite in the request.
   ///
   /// ```
-  /// var invite = await channel.createInvite(maxUses: 2137);
+  /// var invite = await channel.createInvite(maxUses: 100);
   /// ```
```
l7ssha, line 192: docs needs to be only on interface. Doesn't make sense to update them on class that is not shown to end user. Can you remove all docs except this one that are on interfaces?

lib/src/core/channel/guild/guild_channel.dart
```diff
@@ -50,10 +50,24 @@ abstract class IGuildChannel implements IMinimalGuildChannel {
   /// Throws if [entity] isn't [User] or [Role]
   Future<void> deleteChannelPermission(SnowflakeEntity entity, {String? auditReason});
 
-  /// Creates new [IInvite] for [IChannel] and returns it"s instance
+  /// Creates new [IInvite] for [IChannel] and returns it's instance.
+  ///
+  /// Requires the `CREATE_INSTANT_INVITE` permission.
```
Rapougnac, line 55:
```suggestion
  /// Requires the [PermissionsConstants.createInstantInvite] permission.
```
It would be better to link directly to the permission of the lib instead of the api

Rapougnac (changes requested):

l7ssha (approved):

PR #452, opened 2023-03-18
Release 5.0.1

l7ssha:


pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: nyxx
-version: 5.0.0
+version: 5.0.1
```
abitofevrything, line 2: Need to bump version in constants.dart as well

abitofevrything (changes requested):

pubspec.yaml
abitofevrything: Oh, it already is. Missed that.

l7ssha: it is

PR #469, opened 2023-04-03
Rewrite http channels API

### LINKED ISSUE #457: [rewrite] Implement HTTP Channels API
Ref: https://discord.com/developers/docs/resources/channel

abitofevrything:
# Description

Reimplement the HTTP channels API as part of the API.

Closes #457 

## Type of change

# Checklist:


abitofevrything:
How to we want to handle access to methods on channels?

What I mean by this is where do we want to put the convenience methods for channels that would normally go on the `PartialChannel` class so we can do e.g `client.channels[0987654321].sendMessage(...)` without fetching the channel.

Currently I've put all these methods on `PartialChannel`, but the issue channels have that other entities don't is that some methods only work with certain types of channels. For example, `sendMessage` will only work on a text channel, but we have no way of knowing what type of channel a channel is having only its ID.

This isn't really an issue for `sendMessage` (we could just do `client.channels.sendMessage(0987654321, ...)` instead) but it is an issue when the `MessageManager` gets involved. What used to be `client.channels[0987654321].messages.getPins()` would be `MessageManager(client.options.messageCacheConfig, client, 0987654321).getPins()`.

We don't really get any safety out of _not_ putting them on the manager and putting them on concrete channel subclasses either, as users can 1) just cast or 2) use the very long version from above, so we don't really gain anything...

Other libraries (discord.js, discord.py) do seem to opt for putting the methods only on concrete channel subtypes though, so maybe that is the way to go.

What do you think?

abitofevrything:
Since channels are all cached by the client (by default), it actually makes sense to put them on concrete subclasses.

This avoids them polluting the autocomplete of channel types they don't work on anyway.

lib/src/builders/message/message.dart
```diff
@@ -0,0 +1,232 @@
...
+  final List<Embed>? embeds;
+
+  final AllowedMentions? allowedMentions;
+
+  final Snowflake? replyId;
+
+  final bool? requireReplyToExist;
+
+  // TODO
+  //final List<ComponentBuilder>?> components;
```
l7ssha, line 25: when these gonna be done? Also should we move part of interactions library to main framework like Message builder stuff for creating messages with components and needed data structure? All logic for handling interactions should still be in _interactions package but data and parsing part would be better here

l7ssha (approved):

PR #470, opened 2023-04-03
documentation: Guild members

CoocooFroggy:
# Description

Improves documentation of guild member related endpoints. Changes `IGuild#searchMembersGateway` behavior to be consistent with documentation and `IGuild#searchMembers`.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

# Checklist:

### DIFF
--- a/lib/src/core/guild/guild.dart
+++ b/lib/src/core/guild/guild.dart
@@ -309,9 +309,9 @@ abstract class IGuild implements SnowflakeEntity {
   /// Fetches member from API
   Future<IMember> fetchMember(Snowflake memberId);
 
-  /// Allows to fetch guild members. In future will be restricted with `Privileged Intents`.
+  /// Allows to fetch guild members. Restricted with [GatewayIntents.guildMembers].
   /// [after] is used to continue from specified user id.
-  /// By default limits to one user - use [limit] parameter to change that behavior.
+  /// By default limits to one user—use [limit] parameter to change that behavior.
   Stream<IMember> fetchMembers({int limit = 1, Snowflake? after});
 
   /// Returns a [Stream] of [Member]s objects whose username or nickname starts with a provided string.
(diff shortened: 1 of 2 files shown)


lib/src/core/guild/guild.dart
```diff
@@ -309,9 +309,9 @@ abstract class IGuild implements SnowflakeEntity {
   /// Fetches member from API
   Future<IMember> fetchMember(Snowflake memberId);
 
-  /// Allows to fetch guild members. In future will be restricted with `Privileged Intents`.
+  /// Allows to fetch guild members. Restricted with [GatewayIntents.guildMembers].
   /// [after] is used to continue from specified user id.
-  /// By default limits to one user - use [limit] parameter to change that behavior.
+  /// By default limits to one user—use [limit] parameter to change that behavior.
```
abitofevrything, line 314: This shouldn't be changed - the hyphen is used to indicate a change of topic just like in this sentence. Removing the space around it makes it seem like `user-use` is a hyphenated word, which it isn't.

abitofevrything (changes requested):

lib/src/core/guild/guild.dart
CoocooFroggy: Ah I actually changed the hyphen `-` to an em dash `—`. A hyphen is used to combine two words, while an em dash indicates the change in topic.

Note that there are never spaces before or after either of these punctuation marks.

> Hyphens (-) are used to connect two or more words (and numbers) into a single concept, especially for building adjectives. 

https://owl.purdue.edu/owl/general_writing/punctuation/hyphen_use.html

> Dashes are used to set off or emphasize the content enclosed within dashes or the content that follows a dash.

https://owl.purdue.edu/owl/general_writing/punctuation/punctuation/index.html

CoocooFroggy: It's tough to see with a monospace font, but it's more obvious in viewing documentation.

doc

abitofevrything: Right. I'd rather keep this in simple ASCII characters though, where "internet convention" says we use the spacing to indicate the change in topic.

In new documentation I wouldn't mind this, but I see no point in changing preexisting documentation for this.

abitofevrything: Though this is only documentation, so I won't complain too much if you'd rather keep it as is.

CoocooFroggy: Understandable. I think you're right in keeping formatting consistent.

Don't mind me if I make a pull request changing every hyphen to a dash in the future 😋

abitofevrything (approved):

PR #476, opened 2023-04-11
Revert #475

Rapougnac:
This reverts commit 20c1bc8.

# Description
@l7ssha stated that we don't need nullable parameters for cdns.
I also don't see why it should be nullable, if you have a null value convert it to `false` accordingly.

This also removes operations that are unneeded because it's already handled in cdn endpoints.

#475 

## Type of change

# Checklist:

### DIFF
--- a/lib/src/core/application/app_team_user.dart
+++ b/lib/src/core/application/app_team_user.dart
@@ -47,7 +47,7 @@ class AppTeamUser extends SnowflakeEntity implements IAppTeamUser {
   }
 
   @override
-  String avatarUrl({String format = 'webp', int? size, bool animated = false}) {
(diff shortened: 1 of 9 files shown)


abitofevrything:
@l7ssha what's your reasoning here? It seems like a good idea to have an "auto-detect" mode for CDN endpoints - specifically when the URL isn't handled by the code fetching the url

l7ssha:
Autodetect is all fine. I think we should default to animated=true since its natural that if image is animated it should return it as one. I would set all defaults for animated to true

abitofevrything:
Right. @Rapougnac could you add changing the defaults to true? Then we can maintain the behaviour.

abitofevrything (approved):

PR #478, opened 2023-04-30
Rewrite http webhook api

### LINKED ISSUE #466: [rewrite] Implement HTTP Webhook API
Ref: https://discord.com/developers/docs/resources/webhook

abitofevrything:
# Description

Implement the HTTP Webhook resource API

Prerequisite: #469

Closes #466

## Type of change

# Checklist:


lib/src/builders/sentinels.dart
```diff
@@ -0,0 +1,54 @@
...
+  void noSuchMethod(Invocation invocation) => super.noSuchMethod(invocation);
+}
+
+const sentinelImageBuilder = _SentinelImageBuilder();
+
+class _SentinelImageBuilder implements ImageBuilder {
+  const _SentinelImageBuilder();
+
+  @override
+  void noSuchMethod(Invocation invocation) => super.noSuchMethod(invocation);
```
l7ssha, line 53: why it needs to be implemented?

abitofevrything: Sentinels exist to be markers for a given type and not carry any implementation.

We could implement every member to throw `UnimplementedError` or we can use a Dart feature that if a class implements `noSuchMethod`, then it does not need to implement any other members.

This leads to shorter code with ultimately the same result, as no method should ever be called on a sentinel instance.

lib/src/http/managers/webhook_manager.dart
```diff
@@ -0,0 +1,293 @@
...
+      ..messages(id: messageId.toString());
+    final request = BasicRequest(
+      route,
+      method: 'DELETE',
+      queryParameters: {if (threadId != null) 'thread_id': threadId.toString()},
+      authenticated: false,
+    );
+
+    await client.httpHandler.executeSafe(request);
+    // TODO: Can we delete the message from the cache?
```
l7ssha, line 291: I think message should be deleted here, yes

l7ssha (approved):

lib/src/http/managers/webhook_manager.dart
abitofevrything: Yeah - the issue is we don't have the channel id to get the correct cache to remove it from

l7ssha: Then let's leave it for later. Doesn't matter that much I think

abitofevrything: We'll probably also get a gatetway event for the message deletion, so we can handle it there.

PR #483, opened 2023-06-03
Fix invalid casts when fetching channels from cache

abitofevrything:
# Description
Channels in the cache can sometimes be of a different type from the actual channel type, notably when converting text/announcement channels between each other or when using nyxx_interactions, which populates some entries in the cache with instances of `PartialChannel`.

This leads to issues when using `getFromCache`, as although the cache entry does exist, it isn't the correct type.

This PR changes casts that might fail to do extra checks.

## Type of change

# Checklist:


abitofevrything:
Added commits from dev that would have been 5.0.4 before they were reverted.

l7ssha (approved):

PR #485, opened 2023-06-10
Add PartialTextChannel

abitofevrything:
# Description

Adds a `PartialTextChannel` type which allows users to access a channel's mesages without fetching it or creating the `MessageManager` themselves.

## Type of change

# Checklist:

### DIFF
--- a/lib/src/http/managers/channel_manager.dart
+++ b/lib/src/http/managers/channel_manager.dart
@@ -9,6 +9,7 @@ import 'package:nyxx/src/http/request.dart';
 import 'package:nyxx/src/http/route.dart';
 import 'package:nyxx/src/models/channel/channel.dart';
 import 'package:nyxx/src/models/channel/followed_channel.dart';
+import 'package:nyxx/src/models/channel/text_channel.dart';
 import 'package:nyxx/src/models/channel/thread.dart';
 import 'package:nyxx/src/models/channel/thread_list.dart';
 import 'package:nyxx/src/models/channel/types/announcement_thread.dart';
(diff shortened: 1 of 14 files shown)


lib/src/http/managers/channel_manager.dart
```diff
@@ -35,8 +36,18 @@ class ChannelManager extends ReadOnlyManager<Channel> {
...
+  /// Return a partial instance of the entity with ID [id] containing no data.
+  ///
+  /// This allows performing API operations without fetching an instance from the API.
+  ///
+  /// Because this method doesn't perform any API checks, there might be no real entity with the
+  /// correct [id]. In this case, the object returned may not work with the API correctly.
+  ///
+  /// While this method's return type is [PartialChannel], a [PartialTextChannel] is always returned.
+  /// If you are sure the channel you are requesting is a text channel, the returned value can safely
+  /// be cast to a [PartialTextChannel] to access the channel's messages.
```
l7ssha, line 48: this info is kinda contradicting itself. You say that PartialTextChannel is alwas returned and next that if dev is sure that they have text channel they can cast. 

first and foremost why it always return PartialTextChannel?  its gonna change? if not we should remove that docs info since its not relevant

lib/src/models/channel/types/announcement_thread.dart
```diff
@@ -1,17 +1,12 @@
...
-import 'package:nyxx/src/models/message/message.dart';
 import 'package:nyxx/src/models/permission_overwrite.dart';
 import 'package:nyxx/src/models/snowflake.dart';
 import 'package:nyxx/src/models/webhook.dart';
 
-class AnnouncementThread extends Channel implements Thread {
-  @override
-  MessageManager get messages => MessageManager(manager.client.options.messageCacheConfig, manager.client, channelId: id);
-
+class AnnouncementThread extends TextChannel implements Thread {
```
l7ssha, line 9: What do you think abount changing name to `AnnouncementThreadChannel` -- since its channel

l7ssha (approved):

lib/src/http/managers/channel_manager.dart
abitofevrything: Since this is a partial, we can't be sure if the corresponding channel in the API is actually a particular type of channel - so `PartialChannel` indicates a partial channel with an unknown type.

However, it's often the case you want to send a message to a text channel without fetching it to cast it to `TextChannel` - so `PartialTextChannel` is added to mitigate this.

However, how does the developer obtain a `PartialTextChannel`? They could construct it themselves by calling the constructor (which isn't any better than the current solution of creating a `MessageManager`) or we could make the `ChannelManager.operator[]` return a `PartialTextChannel`.

We keep the return type as `PartialChannel` so developers are aware that the returned channel might be any channel type, but we allow them to cast it to a `PartialTextChannel` if they want to. Doing `client.channels[id] as PartialTextChannel` is the developer confirming they know the channel might not be the type they except. The only different with a normal cast is that errors don't occur at the cast itself, but rather later when interacting with the channel.

The difference between the return type and the actual type of returned objects is to force developers to either work with a safe, non-precise type (`PartialChannel`), or manually cast the returned object and accept that it is unsafe.

lib/src/models/channel/types/announcement_thread.dart
abitofevrything: The suffix `Thread` already implies it's a channel, and this terminology (`Thread` vs `Channel`) is used both for other channel types e.g `PublicThread` and `PrivateThread` and also in Discord's documentation.

PR #494, opened 2023-06-14
Change Flags & Snowflake to use BigInts

abitofevrything:
# Description

This PR changes Flags and Snowflakes to use BigInts instead of ints as their "backing store". Their API is unchanged (other than places the value is directly exposed).

This change has a few benefits: we are guaranteed to never overflow when parsing flags or snowflakes (the `Permissions` flags can already go as high as `1 << 42`, and `Snowflake` will run into issues once snowflake values exceed `2 ^ 63`) and this clears the way for (potential) support of the web platform in the future, where the maximum integer limit is much lower (`2 ^ 53`).

## Type of change

# Checklist:

### DIFF
--- a/example/example.dart
+++ b/example/example.dart
@@ -20,7 +20,7 @@ void main() async {
   // Support for partial objects (at minimum with only an ID). These can be accessed for an
   // arbitrary id on a manager with the `[]` operator and allow API operations without fetching the
   // entire object from the API.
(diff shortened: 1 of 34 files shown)


abitofevrything:
Closing this in favour of adding back this functionality if/when needed. Will open a second PR for the small change to the _FlagIterator.

PR #498, opened 2023-06-20
Todos

abitofevrything:
# Description

Fixes most of the TODOs left behind during the rewrite process.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

- [x] Ran `dart analyze` or `make analyze` and fixed all issues
- [x] Ran `dart format --set-exit-if-changed -l 160 ./lib` or `make format` and fixed all issues
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked my changes haven't lowered code coverage

PR #500, opened 2023-06-21
Add getters to get a partial object from an ID where possible.

abitofevrything:
# Description

As the title says.

This means that code can be more fluid:
```dart
// Before
client.onMessageDelete.listen((event) async {
  final channel = client.channels[event.channelId] as PartialTextChannel;
  channel.sendMessage(MessageBuilder(content: 'Message deleted: ${event.id}'));
});

// After
client.onMessageDelete.listen(
  (event) => event.channel.sendMessage(MessageBuild(content: 'Message deleted: ${event.id}')),
);

```

Only to be merged after #498.

## Type of change

# Checklist:

### DIFF
--- a/lib/src/models/gateway/events/invite.dart
+++ b/lib/src/models/gateway/events/invite.dart
@@ -1,5 +1,4 @@
 import 'package:nyxx/src/models/gateway/event.dart';
-import 'package:nyxx/src/models/invite/invite.dart';
 import 'package:nyxx/src/models/invite/invite_metadata.dart';
 import 'package:nyxx/src/models/snowflake.dart';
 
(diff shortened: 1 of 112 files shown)


lib/src/models/gateway/events/invite.dart
```diff
@@ -27,5 +29,11 @@ class InviteDeleteEvent extends DispatchEvent {
   final String code;
 
   /// {@macro invite_delete_event}
-  InviteDeleteEvent({required this.channelId, required this.guildId, required this.code});
+  InviteDeleteEvent({required super.gateway, required this.channelId, required this.guildId, required this.code});
+
+  /// The channel the invite was for.
+  PartialChannel get channel => gateway.client.channels[channelId];
```
l7ssha, line 35: you passed to every event gatewa, but is there a need for gateway? Could it be just client?

l7ssha (approved):

lib/src/models/gateway/events/invite.dart
abitofevrything: The event objects don't have a reference to the client, only a reference to the gateway that than can reference the client. We could add a client getter but I don't think we need to shorten gateway.client to just client.

l7ssha: I was thinking about just passing client ot the event because I presume event doesnt need gateway specifically and its there just to get client. It would make leaner dependencies and easier mocking/extending in future if needed

abitofevrything: It's more consistent to pass gateway because Gateway is essentially the manager for all events - for all other structures we pass the manager so this is "the same" here.

l7ssha: aight you got me

PR #503, opened 2023-06-25
Change PartialTextChannel to return TextChannel when fetching

abitofevrything:
# Description

Change the `get()` and `fetch()` methods on `PartialTextChannel` to return `TextChannel` instead of just `Channel`.

## Type of change

# Checklist:

### DIFF
--- a/lib/src/http/managers/emoji_manager.dart
+++ b/lib/src/http/managers/emoji_manager.dart
@@ -54,7 +54,7 @@ class EmojiManager extends Manager<Emoji> {
   }
 
   @override
-  FutureOr<GuildEmoji> get(Snowflake id) async => await super.get(id) as GuildEmoji;
+  Future<GuildEmoji> get(Snowflake id) async => await super.get(id) as GuildEmoji;
 
   @override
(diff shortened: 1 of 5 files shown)


lib/src/http/managers/emoji_manager.dart
```diff
@@ -54,7 +54,7 @@ class EmojiManager extends Manager<Emoji> {
   }
 
   @override
-  FutureOr<GuildEmoji> get(Snowflake id) async => await super.get(id) as GuildEmoji;
+  Future<GuildEmoji> get(Snowflake id) async => await super.get(id) as GuildEmoji;
```
l7ssha, line 57: why such change?

l7ssha (changes requested):

lib/src/http/managers/emoji_manager.dart
abitofevrything: With how they're implemented, all get() methods always return Futures. We should reflect this in the return type so that users can pass them to APIs which need Futures (e.g Future.wait) without needing to wrap them in Future.value or cast them.

l7ssha (approved):

PR #508, opened 2023-07-03
Rewrite cdn

abitofevrything:
# Description

Adds CDN support to the rewrite.

## Type of change

# Checklist:


lib/src/models/webhook.dart
```diff
@@ -87,6 +89,15 @@ class WebhookAuthor extends PartialWebhook implements MessageAuthor {
...
   WebhookAuthor({required super.id, required super.manager, required this.avatarHash, required this.username});
+
+  @override
+  CdnAsset? get avatar => avatarHash == null
+      ? null
+      : CdnAsset(
+          client: manager.client,
+          base: HttpRoute()..avatars(id: id.toString()),
+          hash: avatarHash!,
+        );
```
l7ssha, line 100: should these things be built in models? would be better to leave building CdnAsset out of here

abitofevrything: Not really, `hash` is part of the model so it can't be extracted. We could move `client` to the manager but then we'd have to have an `avatarFor(Snowflake id, String avatarHash)` method on the manager which feels weird (having to pass an object's property to the manager).

The only thing we could really move out is the `base` route, and I can't think of a place to put that either - the code is already shortened a lot by the extension which allows us to use `..avatars()` instead of `..add(HttpRoutePart(...))`

abitofevrything: It's similar to the partial getters (e.g `message.channel` instead of `client.channels[message.channelId]`) in that it's a convenient shortcut for something the user would otherwise have to type out.

lib/src/http/cdn/cdn_asset.dart
```diff
@@ -0,0 +1,110 @@
...
+  const CdnFormat._(this.extension);
+
+  @override
+  String toString() => 'CdnFormat($extension)';
+}
+
+/// {@template cdn_asset}
+/// An asset, most commonly an image, on Discord's CDN.
+/// {@endtemplate}
+class CdnAsset {
```
l7ssha, line 41: can you add method that created just an url for an asset? That it could be sent that way via message?

lib/src/models/webhook.dart
l7ssha: okay

lib/src/http/cdn/cdn_asset.dart
abitofevrything: There already is - see the `url` getter below.

l7ssha (approved):

lib/src/http/cdn/cdn_asset.dart
l7ssha: im blind, all right the

PR #510, opened 2023-07-04
Add more bitwise operators on `Flags`

Rapougnac:
# Description
Add more bitwise operators on `Flags`

## Type of change

# Checklist:

### DIFF
--- a/lib/src/utils/flags.dart
+++ b/lib/src/utils/flags.dart
@@ -22,7 +22,7 @@ class Flags<T extends Flags<T>> extends IterableBase<Flag<T>> with ToStringHelpe
   /// Return a set of flags that has all the flags set in both `this` and [other].
   Flags<T> operator &(Flags<T> other) => Flags(value & other.value);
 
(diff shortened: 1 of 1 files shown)


lib/src/utils/flags.dart
```diff
@@ -22,6 +22,12 @@ class Flags<T extends Flags<T>> extends IterableBase<Flag<T>> with ToStringHelpe
   /// Return a set of flags that has all the flags set in both `this` and [other].
   Flags<T> operator &(Flags<T> other) => Flags(value & other.value);
 
+  /// Return a set of flags that has all the flags set in `this` but not in [other].
```
abitofevrything, line 25: That's not the correct descriription of XOR - it should be something like this:
```suggestion
  /// Return a set of flags that has all the flags set in `this` or in `other` but not in both.
```

abitofevrything (changes requested):

l7ssha:
Also good addtion would be tests if we have for the rest of operators

PR #548, opened 2023-09-16
Fix example in readme

### LINKED ISSUE #540: Incorrect method usage in README example
The README example uses `connectWebsocket`. That method doesn't exist.

It should be `connectGateway`.

abitofevrything:
# Description

Closes #540.


abitofevrything:
Merged in #549.

Issue #554, opened 2023-09-18
NyxxPluginState is not exported


abitofevrything:
**Describe the bug**
See title.

**To Reproduce**
N/A

**Expected behavior**
The class should be exported.

**Desktop (please complete the following information):**
 - OS: Linux
 - Dart version: 3.1.2
 - Nyxx version: 6.0.0-dev.3

PR #561, opened 2023-09-25
fix: invalid client voice state update, export NyxxPluginState

priyanuj-gogoi:
# Description

Fixes an issue where the client can't mute/deafen itself in a voice channel due to incorrect fields (`mute` & `deaf`) being sent.

The correct fields should be `self_mute` & `self_deaf` as shown in Discord [docs](https://discord.com/developers/docs/topics/voice-connections#retrieving-voice-server-information-gateway-voice-state-update-example).

2nd commit is related to #554 

## Type of change

# Checklist:

### DIFF
--- a/lib/src/builders/voice.dart
+++ b/lib/src/builders/voice.dart
@@ -41,7 +41,7 @@ class GatewayVoiceStateBuilder extends CreateBuilder<VoiceState> {
   @override
   Map<String, Object?> build() => {
         'channel_id': channelId?.toString(),
-        'mute': isMuted,
-        'deaf': isDeafened,
+        'self_mute': isMuted,
+        'self_deaf': isDeafened,
       };
 }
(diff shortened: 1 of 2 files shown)


l7ssha (approved):

lib/src/builders/voice.dart
```diff
@@ -32,16 +32,16 @@ class CurrentUserVoiceStateUpdateBuilder extends VoiceStateUpdateBuilder {
 class GatewayVoiceStateBuilder extends CreateBuilder<VoiceState> {
   Snowflake? channelId;
 
-  bool isMuted;
+  bool isSelfMuted;
 
-  bool isDeafened;
+  bool isSelfDeafened;
```
abitofevrything, line 37: Id rather we didn't rename these fields - isMuted and isDeafened are more consistent, and avoid a breaking change (even if we can make breaking changes before 6.0.0)

abitofevrything (changes requested):

lib/src/builders/voice.dart
priyanuj-gogoi: Sure, I'll revert them back. 😄

abitofevrything (approved):

PR #563, opened 2023-10-01
feature: http interactions preparations

l7ssha:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

Use smart commits here to manipulate issues (eg. Fixes #issue)

If changes in PR are connected to other issues or are affecting code in other parts of framework
(e.g. in main package or any other subpackage) make sure to link issue/PR and describe said problem 

## Type of change

# Checklist:


lib/src/client.dart
```diff
@@ -132,7 +132,7 @@ abstract class Nyxx {
 }
 
 /// A client that can make requests to the HTTP API and is authenticated with a bot token.
-class NyxxRest with ManagerMixin implements Nyxx {
+class NyxxRest with ManagerMixin, EventRestMixin implements Nyxx {
```
abitofevrything, line 135: I'd rather not all `NyxxRest` clients have the `onInteractionCreate` stream as most HTTP only clients will not be receiving the events.

If the events are only emitted if a plugin is added to the client, then the stream should be on the plugin itself.

lib/src/client.dart
```diff
@@ -180,7 +180,7 @@ class NyxxRest with ManagerMixin implements Nyxx {
   }
 }
 
-class NyxxOAuth2 with ManagerMixin implements NyxxRest {
+class NyxxOAuth2 with ManagerMixin, EventRestMixin implements NyxxRest {
```
abitofevrything, line 183: The above comment also solves the issue here as OAuth2 clients _cannot_ receive interaction events (and also cannot connect to the gateway, which is why NyxxOAuth2 doesn't implement NyxxGateway).

lib/src/event_mixin.dart
```diff
@@ -18,10 +19,21 @@ import 'package:nyxx/src/models/gateway/events/webhook.dart';
 import 'package:nyxx/src/models/interaction.dart';
 import 'package:nyxx/src/utils/iterable_extension.dart';
 
+mixin EventRestMixin implements Nyxx {
+  @internal
+  StreamController<Event> onEventController = StreamController.broadcast();
```
abitofevrything, line 24: We should avoid exposing controllers as part of the API. Events should be added to a stream in the same place the stream is created, otherwise the code becomes hard to follow.

Moving the stream to the plugin avoids this problem.

lib/src/gateway/shard.dart
```diff
@@ -52,7 +52,7 @@ class Shard extends Stream<ShardMessage> implements StreamSink<GatewayMessage> {
       } else if (message is EventReceived) {
         final event = message.event;
 
-        if (event is! RawDispatchEvent) {
+        if (event is GatewayEvent) {
```
abitofevrything, line 55: Changing this line modifies the logic so it is now incorrect - `event` is always a `GatewayEvent`, so the `else` branch that outputs more information for dispatch events is always skipped, which reduces verbose logging output.

lib/src/gateway/message.dart
```diff
@@ -34,7 +34,7 @@ abstract class ShardMessage with ToStringHelper {}
 /// A shard message sent when an event is received on the Gateway.
 class EventReceived extends ShardMessage {
   /// The event that was received.
-  final GatewayEvent event;
+  final Event event;
```
abitofevrything, line 37: Shards only ever emit `GatewayEvent`s, so this can be changed back.

lib/src/models/gateway/event.dart
```diff
@@ -5,11 +5,12 @@ import 'package:nyxx/src/utils/to_string_helper/to_string_helper.dart';
 /// {@template gateway_event}
 /// The base class for all events received from the Gateway.
 /// {@endtemplate}
-abstract class GatewayEvent with ToStringHelper {
+abstract class Event with ToStringHelper {}
+
+abstract class GatewayEvent extends Event {
```
abitofevrything, line 10:
```suggestion
/// The base class for all events received by a client.
abstract class Event with ToStringHelper {}

/// {@template gateway_event}
/// The base class for all events received from the Gateway.
/// {@endtemplate}
abstract class GatewayEvent extends Event {
````

File: lib/src/models/gateway/event.dart [hunk 7]
```diff
@@ -5,11 +5,12 @@ import 'package:nyxx/src/utils/to_string_helper/to_string_helper.dart';
...
   /// The opcode of this event.
   final Opcode opcode;
 
-  /// {@macro gateway_event}
```
abitofevrything, line 12: Add this line back once the documentation is fixed.

lib/src/models/gateway/events/interaction.dart
```diff
@@ -4,10 +4,10 @@ import 'package:nyxx/src/models/interaction.dart';
 /// {@template interaction_create_event}
 /// Emitted when an interaction is received by the client.
 /// {@endtemplate}
-class InteractionCreateEvent<T extends Interaction<dynamic>> extends DispatchEvent {
+class InteractionCreateEvent<T extends Interaction<dynamic>> extends Event {
```
abitofevrything, line 7: There isn't a need for this change, a cleaner solution would simply be to have another type of interaction event emitted when using HTTP interactions.

For example, `InteractionCreateEvent` would only be emitted over the gateway, and `WebhookInteractionCreateEvent` (which would not extend `GatewayEvent`) would be emitted when using HTTP interactions.

Both events would simply contain an `interaction` field (and could both implement a common interface), but keeping `InteractionCreateEvent` as a `GatewayEvent` allows for tighter typing in a lot of places (e.g `Gateway.onEvent` and `EventMixin.onEvent`) and allows users to access the client associated with an interaction using `event.gateway.client`,

abitofevrything (changes requested):

lib/src/client.dart
l7ssha: yeah,  stream could be added to plugin but then implementation for nyxx commands would be hard or impossible. We have to have some kind of support in main Nyxx interfaces. Also most client might not receive events but it's supported feature - I dont see problem with exposing possible feature even if rarely it would be used.

PR #574, opened 2023-10-28
fix: initial voice state cache sets null-guild based cache key

priyanuj-gogoi:
## Description

Assume, a bot has connected to gateway or joined a new guild. When the bot receives `GUILD_CREATE` event, internally nyxx will cache & transmit the event data to their respective cache instances as we expect it to do but there is a catch with how nyxx stores the cache for users who are pre-connected into a voice channel before the bot connects to gateway/joins the guild.

https://github.com/nyxx-discord/nyxx/blob/31ab26d4eb1709bb36284029f29be6f13ec0b83b/lib/src/gateway/gateway.dart#L131

In the above code, nyxx transforms `List<VoiceState>` as `MappedListIterable<VoiceState, Map<Snowflake, VoiceState>>` & directly adds them into `Cache<VoiceState>`.

Now, here is the main issue. In `GUILD_CREATE` event, Discord doesn't provide `guild_id` key in `voice_states` (since we already have `event.id` provided). So, the value of `guild_id` here will be `null`. Therefore, from the above code snippet, the `e.cacheKey` value will be of `Snowflake(Object.hash(null, USER_ID))` return value; which is a bit of incorrect. Instead of using `null`, it should use `event.guild.id` value.

Having `null` will not synchronize with the original pre-connected user's cached voice state when the bot receives `VOICE_STATE_UPDATE` event. It will create a new cache entry.

This pull request fixes the initial pre-connected user's voice state cache to synchronize with `VOICE_STATE_UPDATE` event.

### Type of change

A minor breaking change affecting only how nyxx handles voice states caches of pre-connected users.

## Checklist:

[1]: https://pub.dev/documentation/nyxx/latest/nyxx/VoiceState-class.html

### DIFF
--- a/lib/src/gateway/gateway.dart
+++ b/lib/src/gateway/gateway.dart
@@ -42,6 +42,7 @@ import 'package:nyxx/src/models/interaction.dart';
 import 'package:nyxx/src/models/presence.dart';
 import 'package:nyxx/src/models/snowflake.dart';
 import 'package:nyxx/src/models/user/user.dart';
+import 'package:nyxx/src/models/voice/voice_state.dart';
 import 'package:nyxx/src/utils/iterable_extension.dart';
 import 'package:nyxx/src/utils/parsing_helpers.dart';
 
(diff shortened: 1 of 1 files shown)


abitofevrything:
I think a better solution would be to rework how voice states are cached - currently the implementation suffers from cache collisions, and also this problem you've pointed out.

priyanuj-gogoi:
> I think a better solution would be to rework how voice states are cached - currently the implementation suffers from cache collisions, and also this problem you've pointed out.

Could you describe more about the cache collisions?

abitofevrything:
Whoops, typo - hash collisions, not cache collisions. Two different voice states can end up with the same cache key, which means they'll overwrite each other in the cache.

priyanuj-gogoi:
> Whoops, typo - hash collisions, not cache collisions. Two different voice states can end up with the same cache key, which means they'll overwrite each other in the cache.

Hmm, now it makes a bit more sense.

Also, are y'all planning to merge this pull request for at least now to resolve the current cached voice-states issue of pre-connected users until the rework or would prefer to move to the direct rework & fixing this problem as well?

abitofevrything:
I'd just move to the direct rework - either way the fix will be in 4.0.1.

priyanuj-gogoi:
Thank you! 😄

PR #627, opened 2024-01-20
Fixed hoist attribute in both role builder and role update builder

HARUM1122:
The `isHoisted` in `RoleBuilder` and `RoleUpdateBuilder` wasn't working because the key was named 'hoisted but it should be 'hoist'. I've fixed it and tested it as well and now it's working properly.


abitofevrything (approved):

abitofevrything:
Thanks!

PR #657, opened 2024-05-03
Add `guildId` to thread member

Rapougnac:
# Description
Adds the guild id to the thread member for the member manager.
This also adds `withMember` that was not present on fetchThreadMember

## Type of change

# Checklist:

### DIFF
--- a/lib/src/http/managers/channel_manager.dart
+++ b/lib/src/http/managers/channel_manager.dart
@@ -665,7 +665,7 @@ class ChannelManager extends ReadOnlyManager<Channel> {
   }
 
   /// Fetch information about a member in a thread.
-  Future<ThreadMember> fetchThreadMember(Snowflake id, Snowflake memberId, {bool? withMember}) async {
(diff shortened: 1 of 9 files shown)


abitofevrything (commented):
I can see why this is needed. However, if we add this, it'd be inconsistent with other places in the API where we are also unable to determine some ID, and we just fall back to `Snowflake.zero`.

So we need to decide, do we always allow users to specify IDs we are unable to determine ourselves, or do we never allow it?

If we do allow it, then we'd need to add `Snowflake? guildId` parameters to a lot of methods that aren't parsing methods (as most of the times we cannot determine an ID are guild IDs). Is this confusing? You would expect a `guildId` parameter to affect the request, and not the parsing.

Also, if we do add these parameters, how does the user know what ID to pass? If they can determine it, can we determine it (e.g via cache or similar)?

Just a few things to think about.

Rapougnac:
Im dumb :trollface:

PR #665, opened 2024-06-18
Replace (almost) all parsable enums to an enum-like class

### LINKED ISSUE #661: Hang Status is not handled
**Describe the bug**
Discord rolled out new feature called Hang Status for some users. Unfortunately, I wasn't able to find official docs for it, but there are a lot of articles in Google about it.

Apparently, this new status is not handled by [ActivityType](https://github.com/nyxx-discord/nyxx/blob/dev/lib/src/models/presence.dart#L119) and whenever someone sets it, an exception is thrown. Now here comes another problem. After trying many different options to catch this exception, none of them actually worked (I tried adding `onError` callbacks, `handleError` calls, try/catch in different places, `runZoneGuarded`, etc., all this with and without `ignoreExceptions` plugin). Eventually, once this error occurs, bot stops listening to other commands and the only way to fix it is to restart the bot.

**To Reproduce**

- If you have Hang Status feature enabled for your account:
  - Join voice channel and set Hang status;
  - Start bot with `Nyxx.conntectGateway` call, use `GatewayIntents.all`;
  - Observe thrown exception - `FormatException: Unknown activity type`
- If you don't have new Hang status feature enabled for your account:
  - Remove `dnd` key from [UserStatus](https://github.com/nyxx-discord/nyxx/blob/dev/lib/src/models/presence.dart#L28);
  - Set Do not disturb status on the account;
  - Start bot with `Nyxx.conntectGateway` call, use `GatewayIntents.all`;
  - Observe thrown exception - `FormatException: Unknown user status` (which leads to the same issue as described in the issue);

**Expected behavior**
Library correctly handles Hang status for the users.

 - OS: Any;
 - Dart version: 3.3.4
 - Nyxx version: 6.2.1

Rapougnac:
# Description

Closes #661 

## Type of change
- [ ] Breaking change (it's a partial breaking change?)

# Checklist:


l7ssha (approved):

lib/src/utils/enum_like.dart
```diff
@@ -0,0 +1,16 @@
...
+  const EnumLike(this.value);
+
+  @override
+  String toString() => '$runtimeType($value)';
+
+  @override
+  int get hashCode => value.hashCode;
+
+  @override
+  bool operator ==(covariant EnumLike<T, U> other) => identical(this, other) || (other is U && other.value == value);
```
abitofevrything, line 15: The parameter should not be covariant, it should just be `Object`.

abitofevrything (changes requested):
Could you also deprecate all the `parse` constructors? They will probably be removed eventually considering that they're now just duplicate functionality.

abitofevrything (approved):

PR #695, opened 2024-09-06
Add support for zstd compression

abitofevrything:
# Description

Add support for zstd-stream transport compression.

Drafted for now so we can determine whether adding a dependency on another package for this is worth it, especially as it is an unofficial third-party package based on FFI, and not portable.

## Type of change

# Checklist:


lib/src/gateway/shard_runner.dart
```diff
@@ -465,6 +467,9 @@ Stream<dynamic> decompressTransport(Stream<List<int>> raw) {
   });
 }
 
+// See https://github.com/instantiations/es_compression/issues/52 for why the isNotEmpty check is needed.
+Stream<dynamic> decompressZStdTransport(Stream<List<int>> raw) => raw.transform(zstd.decoder).where((message) => message.isNotEmpty);
```
l7ssha, line 471: is isNotEmpty still needed? https://github.com/instantiations/es_compression/issues/52#issuecomment-2336428316

l7ssha (approved):

abitofevrything:
I've preferred to remove our dependency on es_compression as they require you to use their own binary blob which includes a lot more than zstd and is distributed as an opaque blob. Instead we now use zstd_ffi which are just FFI bindings to libzstd, which is now installed by default on ubuntu, debian and dart docker images (but not alpine or the dart `/runtime/` image).

I've also made the switch from JSON+ZLib by default to ETF+ZStd by default, as Discord has shown that ZStd has significant bandwidth improvements and the official client also seems to prefer ETF.

l7ssha (approved):

PR #708, opened 2024-10-07
add support for soundboard

Rapougnac:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

Use smart commits here to manipulate issues (eg. Fixes #issue)

If changes in PR are connected to other issues or are affecting code in other parts of framework
(e.g. in main package or any other subpackage) make sure to link issue/PR and describe said problem 

## Type of change

# Checklist:


lib/src/client_options.dart
```diff
@@ -187,6 +194,8 @@ class RestClientOptions extends ClientOptions {
     this.entitlementConfig = _defaultCacheConfig,
     this.skuConfig = _defaultCacheConfig,
     this.subscriptionConfig = _defaultCacheConfig,
+    this.globalSoundboardCacheConfig = _smallCacheConfig,
+    this.soundboardCacheConfig = _defaultCacheConfig,
```
abitofevrything, line 198: These should be flipped (global => default, guild => small).

lib/src/utils/cache_helpers.dart
```diff
@@ -288,6 +289,7 @@ extension CacheUpdates on NyxxRest {
         EntitlementCreateEvent(:final entitlement) => updateCacheWith(entitlement),
         EntitlementUpdateEvent(:final entitlement) => updateCacheWith(entitlement),
         EntitlementDeleteEvent(:final entitlement) => entitlement.manager.cache.remove(entitlement.id),
+        SoundboardSound(:final user) => updateCacheWith(user),
```
abitofevrything, line 292: The other new events need adding here too (as no-ops) or a warning for unhandled events will be thrown.

lib/src/gateway/gateway.dart
```diff
@@ -1162,6 +1184,37 @@ class Gateway extends GatewayManager with EventParser {
     }
   }
 
+  SoundboardSoundCreateEvent parseSoundboardSoundCreate(Map<String, Object?> raw) {
+    final guildId = maybeParse(raw['guild_id'], Snowflake.parse);
+
+    return SoundboardSoundCreateEvent(
+      gateway: this,
+      sound: client.guilds[guildId ?? Snowflake.zero].soundboard.parse(raw),
```
abitofevrything, line 1192: This shouldn't default to the zero guild's soundboard if guildId is null, it should default to client.sounboard.

Same for other instances of accessing things with `?? Snowflake.zero`.

lib/src/builders/sentinels.dart
```diff
@@ -8,6 +8,8 @@ import 'package:nyxx/src/utils/flags.dart';
 // ASCII encoded "nyxx"
 const sentinelInteger = 0x6E797878;
 
+const sentinelDouble = double.nan;
```
abitofevrything, line 11: It would probably be better to choose a value other than `nan` here (`sentinelInt.toDouble()` would be fine), since nan is more likely to appear as a result of some expression that another relatively large double value.

This is especially true since the current only use of this should be limited to 0-1 (since it's the volume).

abitofevrything (commented):
Maybe add some tests too? We might not be able to test creating/updating sounds, but at least a unit test via `testManager` and an integration test for listing should be possible.

lib/src/client_options.dart
Rapougnac: I don't agree, or at least, put the two configurations at small.
Global sounds don't seem to change frequently, we can suppose they'll never change.
And as of now, there're less than 20 global sounds.

lib/src/builders/sentinels.dart
Rapougnac: `sentinelInt.toDouble()` isn't constant though. That's why I used NaN. Maybe `double.infinity` would be better.
Or any double above 1

lib/src/gateway/gateway.dart
abitofevrything: Alright, fair enough. Didn't realise these events couldn't be triggered by global events.

lib/src/builders/sentinels.dart
abitofevrything: We could use `sentinelInt / 1.0` as a hack to get a constant double.

Infinity has the same problems as NaN, it might arise as the result of some erroneous calculation (in which case we want to include it in the builder payload so the user gets an error instead of ignoring it).

lib/src/builders/soundboard.dart
```diff
@@ -0,0 +1,47 @@
...
+}
+
+class UpdateSoundBuilder extends UpdateBuilder<SoundboardSound> {
+  String name;
+
+  double volume;
+
+  Emoji emoji;
+
+  UpdateSoundBuilder({required this.name, this.volume = sentinelDouble, this.emoji = sentinelEmoji});
```
abitofevrything, line 39: Actually... Why does this need a sentinel if it can't be set to null?

Just make it nullable, and don't include it if it's null.

lib/src/client_options.dart
abitofevrything: The guild one should definitely be small; we don't want many guilds with many stickers to bog down the bot.

The global one can currently be small, but the number of global sounds might grow. As with all caches, we don't care about a few thousand extra entities cached globally - we care about a couple hundred extra cached for every guild the client is in (which, even with just 100 guilds, is 20 000 extra entities).

lib/src/builders/sound.dart
```diff
@@ -0,0 +1,34 @@
...
+  SoundBuilder({required this.data, required this.format});
+
+  static Future<SoundBuilder> fromFile(File file, {String? format}) async {
+    format ??= p.extension(file.path);
+
+    const formats = {
+      'mp3': 'mpeg',
+      'mpeg': 'mpeg',
+      'ogg': 'ogg',
+    };
```
l7ssha, line 18: That could be moved as class or "file const"

abitofevrything: Its only ever used in this function, so it's ok - ImageBuilder does the same

lib/src/builders/soundboard.dart
Rapougnac: It can be set to null, I forgot to include the nullable type

l7ssha (approved):

lib/src/builders/soundboard.dart
```diff
@@ -0,0 +1,47 @@
...
+
+  Emoji? emoji;
+
+  UpdateSoundboardSoundBuilder({required this.name, this.volume = sentinelDouble, this.emoji = sentinelEmoji});
+
+  @override
+  Map<String, Object?> build() => {
+        'name': name,
+        if (volume != sentinelDouble) 'volume': volume,
+        if (emoji != sentinelEmoji) ..._makeEmojiMap(emoji),
```
abitofevrything, line 45: Will error; use `identical(emoji, sentinelEmoji)` instead.

lib/src/models/gateway/events/voice.dart
```diff
@@ -38,3 +42,63 @@ class VoiceServerUpdateEvent extends DispatchEvent {
...
+    required this.userId,
+    required this.emoji,
+    required this.animationType,
+    required this.animationId,
+    required this.soundId,
+    required this.soundVolume,
+  });
+
+  /// The channel this effect was sent in.
+  VoiceChannel get channel => gateway.client.channels[channelId] as VoiceChannel;
```
abitofevrything, line 89: Will error; this should return `PartialChannel`. We currently don't have a way to specify the type of a partial channel (even `PartialTextChannel` only implements `SnowflakeEntity<Channel>`).

lib/src/utils/cache_helpers.dart
```diff
@@ -288,6 +290,14 @@ extension CacheUpdates on NyxxRest {
...
         EntitlementUpdateEvent(:final entitlement) => updateCacheWith(entitlement),
         EntitlementDeleteEvent(:final entitlement) => entitlement.manager.cache.remove(entitlement.id),
+        SoundboardSound() => () {
+            updateCacheWith(entity.user);
+            entity.manager.cache[entity.id] = entity;
+          }(),
+        VoiceChannelEffectSendEvent() => null,
+        SoundboardSoundCreateEvent(:final sound) => updateCacheWith(sound),
+        SoundboardSoundUpdateEvent(:final sound) => updateCacheWith(sound),
+        SoundboardSoundDeleteEvent() => null,
```
abitofevrything, line 300: This should remove the soundboard from the cache

abitofevrything (changes requested):

lib/src/http/managers/soundboard_manager.dart
```diff
@@ -0,0 +1,194 @@
...
+
+    await client.httpHandler.executeSafe(request);
+  }
+}
+
+class GlobalSoundboardManager extends SoundboardManager implements ReadOnlyManager<SoundboardSound> {
+  GlobalSoundboardManager(super.config, super.client) : super(identifier: 'soundboard');
+
+  @override
+  PartialSoundboardSound operator [](Snowflake id) => throw UnsupportedError('Cannot index a global soundboard sound');
```
abitofevrything, line 153: Why?

Rapougnac: Because of `testManager`, since before `GuildSoundboardManager` didn't implement `Manager<T>` but `ReadOnlyManager<T>`.

abitofevrything (approved):

PR #726, opened 2024-10-30
properly censor the token when connecting

Rapougnac:
# Description

title

## Type of change

# Checklist:


abitofevrything (changes requested):
I would prefer to just log the connection information after plugins have run their `beforeConnect` (where the Logging plugin sets up token censoring).

This can be done by moving the logging lines to the inside of the `_doConnect` callback parameter IIRC.

abitofevrything (approved):

PR #740, opened 2025-01-28
Make `SoundboardSound` have a CDN asset

Lexedia:
# Description
title

## Type of change

## Checklist:

### DIFF
--- a/lib/src/models/soundboard/soundboard.dart
+++ b/lib/src/models/soundboard/soundboard.dart
@@ -1,4 +1,6 @@
+import 'package:nyxx/src/http/cdn/cdn_asset.dart';
 import 'package:nyxx/src/http/managers/soundboard_manager.dart';
+import 'package:nyxx/src/http/route.dart';
 import 'package:nyxx/src/models/emoji.dart';
 import 'package:nyxx/src/models/guild/guild.dart';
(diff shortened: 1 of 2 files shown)


lib/src/models/soundboard/soundboard.dart
```diff
@@ -10,6 +12,16 @@ class PartialSoundboardSound extends WritableSnowflakeEntity<SoundboardSound> {
   final SoundboardManager manager;
 
   PartialSoundboardSound({required super.id, required this.manager});
+
+  /// The sound asset for this soundboard sound.
+  CdnAsset get sound => CdnAsset(
+        client: manager.client,
+        base: HttpRoute()..soundboardSounds(),
+        // We can't use .soundboardSounds(id: id.toString()) because hash would be empty, adding an empty string will populate
+        // `route.segments` which will join it with a trailing slash and Discord doesn't handle trailing slashes.
```
abitofevrything, line 21: Can we instead fix HttpRoute to filter out empty segments before joining (i.e not add a trailing slash, since that's technically a bug)?

Alternatively fix CdnAsset to check for an empty hash. Though overall I think we should leave a TODO to make hash nullable in the next major release.

abitofevrything (changes requested):

abitofevrything (approved):

PR #746, opened 2025-03-22
Properly handle unknown union types

abitofevrything:
# Description

See title.


Lexedia (commented):
Once the conflicts are resolved, can you expose the models?

abitofevrything:
Models are now exposed. Conflicts will be resolved locally, since we can't merge this to main directly (the tip of main already has commits for 6.7.0).

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 6.6.1
+__22.03.2025__
+
+- feat: Properly handle unknown union types.
```
Lexedia, line 4: I think this would count as a fix rather than a feature.
but since we don't really care about being consistent it's alright 😄

Lexedia (approved):

PR #748, opened 2025-05-11
Implement interaction callback response

Lexedia:
# Description
Please include a summary of why this pull request should be merged.

### DIFF
--- a/lib/src/http/managers/interaction_manager.dart
+++ b/lib/src/http/managers/interaction_manager.dart
@@ -285,8 +285,44 @@ class InteractionManager {
     );
   }
 
+  InteractionCallbackResponse parseInteractionCallbackResponse(Map<String, Object?> raw) {
+    return InteractionCallbackResponse(
+      interaction: parseInteractionCallback(raw['interaction'] as Map<String, Object?>),
+      resource: maybeParse(raw['resource'], parseInteractionResource),
+    );
+  }
+
+  InteractionCallback parseInteractionCallback(Map<String, Object?> raw) {
+    return InteractionCallback(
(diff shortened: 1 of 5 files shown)


lib/src/http/managers/interaction_manager.dart
```diff
@@ -313,17 +349,25 @@ class InteractionManager {
...
       );
     }
 
-    await client.httpHandler.executeSafe(request);
+    final response = await client.httpHandler.executeSafe(request);
+
+    if (withResponse == null || withResponse == false) {
+      return null;
+    }
+
```
abitofevrything, line 369: Needs client.updateCacheWith call here (for the nested Message object). This will also require updating client.updateCacheWith to handle the appropriate data structures.

abitofevrything (changes requested):

lib/src/http/managers/interaction_manager.dart
Lexedia: Good catch, thanks.

lib/src/models/interaction.dart
```diff
@@ -162,13 +162,15 @@ mixin MessageResponse<T> on Interaction<T> {
       _didRespond = true;
       _wasEphemeral = isEphemeral;
 
-      await manager.createResponse(id, token, InteractionResponseBuilder.channelMessage(builder, isEphemeral: isEphemeral));
+      return manager.createResponse(id, token, InteractionResponseBuilder.channelMessage(builder, isEphemeral: isEphemeral));
```
abitofevrything, line 165: Looks like there should be `withResponse: withResponse` here.

abitofevrything (approved):

PR #750, opened 2025-05-17
Message update has no longer a partial payload

Lexedia:
# Description
See https://github.com/discord/discord-api-docs/pull/7017

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [x] Ran `dart analyze` or `make analyze` and fixed all issues
- [x] Ran `dart format --set-exit-if-changed -l 160 ./lib` or `make format` and fixed all issues

PR #749, opened 2025-05-12
Support audit log for thread creation

Lexedia:
# Description
Please include a summary of why this pull request should be merged.

PR #751, opened 2025-05-30
Release 6.7.0

abitofevrything:
# Description

As the title says.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
 ## 6.7.0
 __30.05.20205__
 
-- See the changelog for 6.7.0-dev.1 for details.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 6.7.0
+__30.05.20205__
+
+- See the changelog for 6.7.0-dev.1 for details.
```
Lexedia, line 4: What about #747, #748, #749 and #750?

Lexedia (changes requested):

CHANGELOG.md
l7ssha: I think we should avoid avoid referencing other versions like that all together. Just copy what ever was release in dev version and if needed add new stuff

l7ssha (approved):

PR #760, opened 2025-07-17
ignore tools and test directories when publishing

Lexedia:
# Description
it doesnt really make sense to have these as a published package, and it also reduces the package size by half.

## Type of change


.pubignore
```diff
@@ -0,0 +1,2 @@
+tools/*
+test/*
```
abitofevrything, line 2: .pubignore overrides .gitignore, so we need to add some of files ignored in there that will be auto-generated by the release workflow. Also added `.github/` to the ignore list.

```suggestion
# See https://www.dartlang.org/guides/libraries/private-files

# Files and directories created by pub
.dart_tool/
.packages
build/
pubspec.lock

# Files not needed by dependents
.github/
tools/
test/
```

abitofevrything (changes requested):

abitofevrything (changes requested):
We need the entirety of .gitignore in here. The current setup is enough for releases done automatically on github, but locally, there may be even more files - for example I have `doc/api` filled out from times I was testing documentation generation.

If these additional entries from .gitignore are not in .pubignore, then if the need for a local publish ever arises (as is common for dev releases), we'd have to pay extra care to get rid of those files first. Not ideal.

PR #773, opened 2026-02-01
fix `redirect_uris` containing null values

Lexedia:
# Description
See https://github.com/discord/discord-api-docs/issues/7586


abitofevrything (changes requested):
Change looks good, but could you also include the necessary changes to release the patch? i.e bump versions in pubspec and ApiOptions, and update changelog.

abitofevrything (approved):

PR #775, opened 2026-04-10
Add topics to documentation

abitofevrything:
As the title says.

The added documentation per-topic should also help to make the package's structure more obvious to newcomers.


Lexedia (approved):
https://github.com/user-attachments/assets/39f179a0-12df-4db6-9922-040fce3e241a

PR #783, opened 2026-05-26
update the readme example

Lexedia:
# Description
dots shorthands my beloved

also Hacktoberfest accepted even tho we're not even in october

i should get a raise because the amout of tokens ive been burning to write those shitposts PRs is kinda crazy.

lovewires

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,13 +15,13 @@ import 'package:nyxx/nyxx.dart';
 void main() async {
   final client = await Nyxx.connectGateway('<TOKEN>', GatewayIntents.allUnprivileged);
 
-  final botUser = await client.users.fetchCurrentUser();
+  final botUser = await client.user.get();
 
   client.onMessageCreate.listen((event) async {
     if (event.mentions.contains(botUser)) {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,13 +15,13 @@ import 'package:nyxx/nyxx.dart';
...
 
-  final botUser = await client.users.fetchCurrentUser();
+  final botUser = await client.user.get();
 
   client.onMessageCreate.listen((event) async {
     if (event.mentions.contains(botUser)) {
       await event.message.channel.sendMessage(MessageBuilder(
         content: 'You mentioned me!',
-        referencedMessage: MessageReferenceBuilder.reply(messageId: event.message.id),
+        referencedMessage: .reply(messageId: event.message.id),
```
abitofevrything, line 24: Nyxx's lower SDK bound is too low for dot shorthands, so this won't work on all SDK versions we support. We should probably leave it as-is for now.

abitofevrything (changes requested):

README.md
Lexedia: Gotta love verbosity 😔

PR #784, opened 2026-05-30
Update documentation links to nyxx.org

abitofevrything:
As the title says.

Also incremented the version build number so the changes get pushed to pub.


l7ssha (approved):

Lexedia:
my goat

PR #787, opened 2026-08-18
Added `final String? name` to `guild.dart` to respect partial guild r…

shalien:
…esponse inside `Invite.fetch()`

# Description

Add a field String? name into PartialGuild to respect actual response from Invite.fetch() / Invite Object according to discord documentation : https://docs.discord.com/developers/resources/invite#invite-object 

Also updated UserGuild to use the super.name argument from parent

## Type of change

## Checklist:


abitofevrything (changes requested):
Nyxx does not expose fields on partials. If you want to include this data, create a specific subclass of PartialGuild for use in invites.

PR #788, opened 2026-09-15
Document where we intentionally return without awaiting from within a try block

abitofevrything:
# Description

A new lint rule now warns against this situation, which is very useful to have. We do have a couple places where we do this intentionally though, so document those locations and ignore the lint there.


l7ssha (approved):

lib/src/http/handler.dart
```diff
@@ -249,6 +249,10 @@ class HttpHandler {
         _realLatencies.removeFirst();
       }
 
+      // We want to remove the in-flight request as soon as the response is received
+      // or the request encounters an error.
+      // The try-finally is intentionally shortcut by returning early.
+      // ignore: unawaited_return_in_try_block
       return _handle(request, response);
```
Lexedia, line 256: Wouldn't `unawaited` from `dart:async` get rid of the lint?
We can still keep the comments.

lib/src/http/handler.dart
```diff
@@ -316,6 +320,9 @@ class HttpHandler {
         _pendingRateLimits.add(completer);
         try {
           await completer.future;
+          // We want to clean up our rate limit timers as soon as the completer ends.
+          // Shortcut the try-finally block by returning without await.
+          // ignore: unawaited_return_in_try_block
           return _execute(request);
```
Lexedia, line 326: Same here

lib/src/http/handler.dart
abitofevrything: Unawaited fully ignores the future, but we want to preserve the return value.

lib/src/http/handler.dart
abitofevrything: Same as above

lib/src/http/handler.dart
Lexedia: i may be stupid
HRubWoaWwAAAyBD

